PR #7, opened 2018-02-27
Macos example app

stuartmorgan:
Adds a sample macOS embedder app. See commit comments for details.


example_flutter/pubspec.lock
```diff
@@ -0,0 +1,398 @@
+# Generated by pub
```
awdavies: pubspec.lock shouldn't be included.  We should at `*.lock` to our `.gitignore` file.

awdavies: I take that back.  Looks good!

PR #11, opened 2018-02-28
Add support for snapshot mode to FLEViewController

### LINKED ISSUE #8: Make the macOS example app self-contained
Currently it references all the Flutter code where it is in the source tree. It should bundle it all, so the example is of a working, self-contained application bundle.

stuartmorgan:
The embedding API allows passing empty main and package paths to use
snapshot mode rather than source mode, which allows for easy bundling.

Switches the example app to use snapshot mode, and bundles the build
resources directory, so that the app is self-contained. Removes the
need for the command-line argument workaround (which was accidentally
missing from the previous patch due to the scheme that passed the path
to the example app directory not being marked as shared, so this has
the added effect of fixing the sample app).

Closes #8


stuartmorgan:
Per the TODO, the API is a bit odd here because I didn't want to churn too much until we know whether supporting source mode is going to be useful in the long term. If we keep both it probably makes more sense to bundle the paths into a config object with two different constructors, and have a single engine-launching call that takes one of those configs.

Also, the diff is confused by the fact that I moved a large method down to the private methods section. The plugin and platform message code isn't changing.

PR #20, opened 2018-03-11
[linux] Do not pass local-engine-src arguments to build command

vially:
Given that we're already copying the `libflutter_engine.so` to the `library` directory, it looks like there's no need to specify the `--local-engine` and `--local-engine-src-path` arguments to the build command.

Also, this allows building the embedder (`libflutter_embedder.so`) without having to configure the host machine with all the tools needed for building the engine (which can be done in a container and then only extracting `libflutter_engine.so`).

### DIFF
--- a/linux/example/Makefile
+++ b/linux/example/Makefile
@@ -12,44 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Before editing these, make sure you have read the documentation on how to
-# build the flutter engine here:
-# https://github.com/flutter/engine/blob/master/CONTRIBUTING.md
-#
-# $(FLUTTER_ENGINE_PATH) is the variable pointing to the git checkout of the
-# flutter engine, under the assumption that you have already built it. Here,
-# the flutter engine path is relative to
-# `flutter-desktop-embedding/example_flutter`, as this Makefile will build that
-# application during the build process.
-#
-# This variable will be used when building `example_flutter/` to pass the
-# `--local-engine-src-path` flag to the flutter binary.
-FLUTTER_ENGINE_PATH=../../engine
-
-# See above regarding building the flutter engine before editing these
-# variables.
-#
-# $(FLUTTER_ENGINE_BUILD) refers to the `gn` configuration used to build the
-# engine. When running `gn --unoptimized` to generate the build configuration,
-# the variable should be set (as it is now) to `host_debug_unopt`. If building
-# with the flags `--runtime-mode release`, then this variable will be set to
-# `host_release`.
-#
-# A simple way to figure out how to set this flag is to check the output of `gn`
-# and then set $(FLUTTER_ENGINE_BUILD) to be the name of the directory after `out/`
-#
-# Example output:
-#
-#```
-# $ ./src/flutter/gn --unoptimized
-# gn gen --check in out/host_debug_unopt
-# Done. Made 385 Targets from 168 files in 379ms
-#```
-#
-# From the above, you should then set this flag to `host_debug_unopt`.
-FLUTTER_ENGINE_BUILD=host_debug_unopt
-
-FLUTTER_ENGINE_SRC=$(FLUTTER_ENGINE_PATH)/src
 FLUTTER_EMBEDDER_LIB=flutter_embedder
 FLUTTER_EXAMPLE_DIR=../../example_flutter
 FLUTTER_BIN_FROM_EXAMPLE_DIR=../../flutter/bin/flutter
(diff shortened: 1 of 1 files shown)


stuartmorgan:
If you run the resulting build output in snapshot mode, aren't you seeing a "Wrong script snapshot version" message in the console? That's the reason those flags are there (see https://github.com/flutter/engine/blob/master/CONTRIBUTING.md#building-and-running-the-code).

Copying to the library directory is for the benefit of the C++ embedder build, not the Flutter commands.

awdavies:
Those flags are in place to prevent an error that is occasionally surfaced when running a Flutter application wherein there is a checksum mismatch due to using a local build of the engine (I don't have the full details of what surfaces the checksum mismatch yet), which, as Stuart stated surfaces a "Wrong script snapshot version."

vially:
I'm very new to flutter so I don't fully understand how everything fits together. 

What I did is I built `libflutter_engine.so` in a container (because building the engine on ArchLinux wasn't very straight forward) and I copied it over to the `flutter-desktop-embedding/linux/library` directory in my host filesystem. Then I was able to build and run the `flutter_embedder_example` without having to specify the `--local-engine` args.


Here's a Dockerfile which sums up what I did (except I did on my host OS filesystem rather than in a container):
```
FROM debian:stretch

WORKDIR /tmp

RUN apt-get update && apt-get install -y build-essential wget curl unzip git libglfw3-dev libglu1 lib32stdc++6

RUN git clone --depth 0 https://github.com/flutter/flutter.git
RUN git clone -b linux-no-local-engine https://github.com/vially/flutter-desktop-embedding.git
RUN wget -O /tmp/flutter-desktop-embedding/linux/library/include/embedder.h https://raw.githubusercontent.com/flutter/engine/2dd385fd7efd2889a1b04a77d9d24c4ebdbb561a/shell/platform/embedder/embedder.h
ADD libflutter_engine.so /tmp/flutter-desktop-embedding/linux/library/
RUN cd /tmp/flutter-desktop-embedding/linux && make
```

You can see by building this `Dockerfile` that the build does actually succeed (you'll need to provide it the `libflutter_engine.so` though).

So I'm not exactly sure how I can trigger the "Wrong script snapshot version" error. Also, I don't know what it means to "run the resulting build output in snapshot mode". I've only tried running the resulting binary directly (e.g.: `./flutter_embedder_example`).

Is this error supposed to be happening at build time or at runtime? I'd love to dig more into it to see what's causing the issue but I'll need some guidance from you in order to trigger the error first.

PS: Great work on this and I'm very excited about the potential of flutter (especially in the context of linux desktop apps).

awdavies:
So when you run `./flutter_embedder_example` it is running a pre-compiled version of the flutter application from within the example directory.  If you look at the `Makefile` output you can see the following:

```
make[1]: Entering directory '/whateveryourflutterrootis/flutter-desktop-embedding/linux/example'
cd ../../example_flutter; \
../../flutter/bin/flutter build flx \
        --local-engine-src-path=../../engine/src \
        --local-engine=host_debug_unopt;
```
This compiles a snapshot of the application using your engine.

The fact that "Wrong script snapshot version" _doesn't_ surface means the engine that was used to compile the flutter binary into a snapshot is the same as what you've built to make `libflutter_engine.so`.

On my machine (and I imagine others), this error is raised when I don't supply the path to the engine and build and I attempt to run the application in snapshot mode.

Snapshot mode just means that the Flutter engine "runs" the application to precompile it into a "snapshot" (imagine something akin to `.pyc` files).

I don't think this change is necessary unless it is somehow breaking your ability to run the `Makefile`, in which case it may be possible to add a variable to track in the Makefile.

stuartmorgan:
With Andrew's comments as background, some specific responses to your questions/comments:

> So I'm not exactly sure how I can trigger the "Wrong script snapshot version" error.

Running the resulting app is when it would show up in the console.

> Also, I don't know what it means to "run the resulting build output in snapshot mode". I've only tried running the resulting binary directly (e.g.: ./flutter_embedder_example).

The example uses snapshot mode, so if you're just running the example without modifications that would do it.

> Is this error supposed to be happening at build time or at runtime?

Runtime.

It's not clear to me at this point if you are actually not seeing the error, or just weren't looking at the right step for it.

vially:
@awdavies @stuartmorgan Thanks for your detailed responses. They are very helpful.

I can still run the example in snapshot mode without seeing the error, but I think I'm starting to understand now why it wouldn't always work.

My initial plan was to find an easy way to package `flutter`, the `engine` and `flutter-desktop-embedding` in a way that they can be installed separately on Archlinux using the distribution's package manager (and it's dependency model). And I was hoping that as long as you have `flutter` and `engine` installed on your system, you wouldn't need this monorepo structure when building `flutter-desktop-embedding`.

But now that you've made things a bit clearer, it seems like what I wanted to do might not be as straight forward as I hoped.

I'll close this PR for now and I'll look into alternative solutions.

PR #22, opened 2018-03-12
#19 linux target g++ command order

code-mc:
Currently linux target gives an undefined reference due to the sources being included at the end of the g++ command. See #19


code-mc:
Should be ok now

code-mc:
Nvm I guess, what a hassle

PR #37, opened 2018-04-04
Use actual pixel ratio on macOS

stuartmorgan:
The pixel ratio was hard-coded to 1.0; this uses the actual
pixel ratio, and enables high-resolution support on the OpenGL view.

macOS portion of Issue #32


stuartmorgan:
Note that most of the changes in the XIB file are just churn from touching it. The important part is enabling wantsBestResolutionOpenGLSurface.

krisgiesing:
LGTM!

PR #39, opened 2018-04-09
[linux] Fix jsoncpp deprecations

vially:
The `jsoncpp` [Reader](http://open-source-parsers.github.io/jsoncpp-docs/doxygen/class_json_1_1_reader.html) and [FastWriter](http://open-source-parsers.github.io/jsoncpp-docs/doxygen/class_json_1_1_fast_writer.html) classes are [depreacted](http://open-source-parsers.github.io/jsoncpp-docs/doxygen/deprecated.html).

Because of this `flutter-desktop-embedding` fails to compile using `jsoncpp`-`v1.8.4` on ArchLinux.

This PR attempts to fix this by using the newer `jsoncpp` API.

PS: Please review this carefully because I don't know `C` and I'm unable to run this code in order to test it. It does seem to compile though.

### DIFF
--- a/linux/library/src/flutter_embedder.cc
+++ b/linux/library/src/flutter_embedder.cc
@@ -81,13 +81,16 @@ static void GLFWOnFlutterPlatformMessage(const FlutterPlatformMessage *message,
     return;
   }
   GLFWwindow *window = reinterpret_cast<GLFWwindow *>(user_data);
-  Json::Reader parser;
+  Json::CharReaderBuilder reader_builder;
+  std::unique_ptr<Json::CharReader> parser(reader_builder.newCharReader());
   Json::Value json;
-  std::string raw_string(reinterpret_cast<const char *>(message->message), 0,
-                         message->message_size);
-  bool parsing_successful = parser.parse(raw_string, json);
+  std::string parse_errors;
+  auto raw_message = reinterpret_cast<const char *>(message->message);
+  bool parsing_successful = parser->parse(
+      raw_message, raw_message + message->message_size, &json, &parse_errors);
   if (!parsing_successful) {
-    std::cerr << "Unable to parse platform message" << std::endl;
+    std::cerr << "Unable to parse platform message" << std::endl
+              << parse_errors << std::endl;
     return;
   }
   auto state = GetSavedEmbedderState(window);
@@ -99,8 +102,8 @@ static void GLFWOnFlutterPlatformMessage(const FlutterPlatformMessage *message,
     return;
   }
 
-  Json::FastWriter writer;
-  std::string output = writer.write(response);
+  Json::StreamWriterBuilder writer_builder;
+  std::string output = Json::writeString(writer_builder, response);
   FlutterPlatformMessage platform_message_response = {
       .struct_size = sizeof(FlutterPlatformMessage),
       .channel = channel.c_str(),


linux/library/src/flutter_embedder.cc
```diff
@@ -81,13 +81,15 @@ static void GLFWOnFlutterPlatformMessage(const FlutterPlatformMessage *message,
...
   GLFWwindow *window = reinterpret_cast<GLFWwindow *>(user_data);
-  Json::Reader parser;
+  Json::CharReaderBuilder rbuilder;
+  std::unique_ptr<Json::CharReader> const parser(rbuilder.newCharReader());
   Json::Value json;
+  std::string parse_errors;
   std::string raw_string(reinterpret_cast<const char *>(message->message), 0,
                          message->message_size);
-  bool parsing_successful = parser.parse(raw_string, json);
+  bool parsing_successful = parser->parse(raw_string.c_str(), raw_string.c_str() + raw_string.size(), &json, &parse_errors);
```
stuartmorgan: This and 92 will need to be wrapped to follow the C++ style guide length limits.

https://google.github.io/styleguide/cppguide.html

But it also doesn't seem like there's anything to be gained by round-tripping through std::string with this API.

linux/library/src/flutter_embedder.cc
```diff
@@ -99,8 +101,8 @@ static void GLFWOnFlutterPlatformMessage(const FlutterPlatformMessage *message,
     return;
   }
 
-  Json::FastWriter writer;
-  std::string output = writer.write(response);
+  Json::StreamWriterBuilder wbuilder;
```
stuartmorgan: 'wbuilder' (and 'rbuilder' above) will similarly need to be changed to meet the style guide guidance on not abbreviating names.

linux/library/src/flutter_embedder.cc
```diff
@@ -99,8 +101,8 @@ static void GLFWOnFlutterPlatformMessage(const FlutterPlatformMessage *message,
...
+  std::string output = Json::writeString(wbuilder, response);
```
stuartmorgan: According to the docs this is creating and destroying a new writer under the hood, much like the code above. If we have to heap-allocate reader/writer objects anyway, maybe we should just cache them? Unless the embedding API sends this on arbitrary threads (which I don't think is the case) it would be pretty easy.

awdavies:
I implemented things this way just from checking out the code from github and building the docs.  The deprecated link didn't mention anything about it, but double-checking the headers it does indicate to use StreamWriterBuilder instead :+1:

linux/library/src/flutter_embedder.cc
awdavies: At the moment everything should be running in the same thread, so it should be fine.

vially:
@awdavies @stuartmorgan I've addressed most of your comments except caching the reader/writer. I can take a stab at that as well if you feel it's important.

awdavies:
Verified it compiles and runs :+1:

linux/library/src/flutter_embedder.cc
```diff
@@ -81,13 +81,16 @@ static void GLFWOnFlutterPlatformMessage(const FlutterPlatformMessage *message,
     return;
   }
   GLFWwindow *window = reinterpret_cast<GLFWwindow *>(user_data);
-  Json::Reader parser;
+  Json::CharReaderBuilder reader_builder;
+  std::unique_ptr<Json::CharReader> const parser(reader_builder.newCharReader());
```
awdavies: Remove the `const` here:

```c++
std::unique_ptr<Json::CharReader> parser(reader_builder.newCharReader());
``` 

Should be fine.

vially: Fixed

PR #42, opened 2018-04-17
[linux] Update build to use `bundle`.

awdavies:
`build flx` is deprecated.  Use `bundle` instead.


awdavies:
Woops!  Another pull request handles both mac and linux.

PR #49, opened 2018-05-02
Add configuration keys and logic to file chooser plugin.

evilBird:
This change updates the file chooser plugin code, adding keys and logic that extend NSSavePanel/NSOpenPanel configuration to include the following options:
*provide an initial string value to display in panel textField (nameFieldStringValue property)
*enable/disable choosing files (canChooseFiles property)
*set title property
*set message property
*set prompt property


macos/library/FLEFileChooserPlugin.m
```diff
@@ -30,11 +30,21 @@
...
-// Method argument keys.
+// Appearance configuration keys for NSSavePanel.
+static NSString *const kPanelTitleKey = @"title";
+static NSString *const kPanelMessageKey = @"message";
+static NSString *const kOKButtonTextKey = @"prompt";
+
+// Method argument keys for NSSavePanel.
+static NSString *const kInitialDirectoryKey = @"initialDirectory";
 static NSString *const kAllowedFileTypesKey = @"allowedFileTypes";
+static NSString *const kFileNameStringValueKey = @"nameFieldStringValue";
```
stuartmorgan: Similarly, something like initialFileName would be clearer here.

macos/library/FLEFileChooserPlugin.m
```diff
@@ -30,11 +30,21 @@
...
+static NSString *const kCanCreateDirectoriesKey = @"canCreateDirectories";
```
stuartmorgan: Do you have a specific use case for needing to configure this? It seems like we could probably just have it be on for save panels, and not add an option.

macos/library/FLEFileChooserPlugin.m
```diff
@@ -30,11 +30,21 @@
 static NSString *const kPlatformMethodNameKey = @"method";
 static NSString *const kPlatformMethodArgsKey = @"args";
 
-// Method argument keys.
+// Appearance configuration keys for NSSavePanel.
+static NSString *const kPanelTitleKey = @"title";
+static NSString *const kPanelMessageKey = @"message";
+static NSString *const kOKButtonTextKey = @"prompt";
```
stuartmorgan: This message name is very specific to NSSavePanel; these plugins are intended to be cross-platform, so should use more descriptive names. confirmButtonLabel maybe?

macos/library/FLEFileChooserPlugin.m
```diff
@@ -96,6 +121,9 @@ - (void)configureOpenPanelForClient:(nonnull NSNumber *)clientID
     ((NSOpenPanel *)_panels[clientID]).canChooseDirectories =
         [arguments[kCanChooseDirectoriesKey] boolValue];
   }
+  if ([argKeys containsObject:kCanChooseFilesKey]) {
+    ((NSOpenPanel *)_panels[clientID]).canChooseFiles = [arguments[kCanChooseFilesKey] boolValue];
```
stuartmorgan: (NSOpenPanel *)_panels[clientID] is being repeated enough times now that it should be extracted to a new local variable at the start of the method.

macos/library/FLEFileChooserPlugin.m
evilBird: Done.

macos/library/FLEFileChooserPlugin.m
evilBird: That's a good point, I don't believe there is a particular case in which we'd want to disable this for a save panel, nor to enable it for an open panel. Done.

macos/library/FLEFileChooserPlugin.m
evilBird: I chose this key name so that its semantics communicate the parameter's association with a specific value of the FileChooserResult enum (FileChooserResult.OK) on the flutter side. That said, I'm not convinced that 'OK button text' is capturing that. So, I've tried to split the difference by changing the key to 'OKButtonLabel'.

macos/library/FLEFileChooserPlugin.m
evilBird: Done.

krisgiesing:
Yes. I'm assuming that plugins in general know how to ignore properties they don't specifically know about? (Otherwise it seems like the dart-side change would have been breaking.) If so then that can be a separate PR, although doing both at once would probably make it easier to review for consistency.

macos/library/FLEFileChooserPlugin.m
```diff
@@ -147,6 +172,7 @@ - (nullable id)handlePlatformMessage:(NSDictionary *)message {
   __weak FLEFileChooserPlugin *weakself = self;
   if ([methodName isEqualToString:kShowSavePanelMethod]) {
     NSSavePanel *savePanel = [NSSavePanel savePanel];
+    savePanel.canCreateDirectories = true;
```
stuartmorgan: s/true/YES/

macos/library/FLEFileChooserPlugin.m
```diff
@@ -30,11 +30,20 @@
 static NSString *const kPlatformMethodNameKey = @"method";
 static NSString *const kPlatformMethodArgsKey = @"args";
 
-// Method argument keys.
+// Appearance configuration keys for NSSavePanel.
+static NSString *const kPanelTitleKey = @"title";
+static NSString *const kPanelMessageKey = @"message";
```
stuartmorgan: Per offline discussion, since NSSavePanel's title no longer shows up in 10.13, and I've never encountered a save/open panel using message (and it looks bad on 10.13, which is more evidence that it's extremely rare), let's remove these. They also aren't easily implementable on Linux.

macos/library/FLEFileChooserPlugin.m
```diff
@@ -30,11 +30,20 @@
...
+// Method argument keys for NSSavePanel.
+static NSString *const kInitialDirectoryKey = @"initialDirectory";
 static NSString *const kAllowedFileTypesKey = @"allowedFileTypes";
+static NSString *const kInitialFileName = @"nameFieldStringValue";
+
+// Method argument keys for NSOpenPanel.
 static NSString *const kAllowsMultipleSelectionKey = @"allowsMultipleSelection";
 static NSString *const kCanChooseDirectoriesKey = @"canChooseDirectories";
-static NSString *const kInitialDirectoryKey = @"initialDirectory";
+static NSString *const kCanChooseFilesKey = @"canChooseFiles";
```
stuartmorgan: Given that for our Linux implementation it would be a hack to allow both files and directories, let's just document the Dart side to make it clear that canChooseDirectories enables directories but disables files (and change the the macOS plugin to match that behavior).

We can rename it later, once the Dart code is in this repo.

If at some point someone has a specific use case for this, we can revisit doing a hack on Linux (adding another button to a file chooser that allows folders seems to be an option).

awdavies: :+1:

macos/library/FLEFileChooserPlugin.m
evilBird: Done.

macos/library/FLEFileChooserPlugin.m
evilBird: Done.

macos/library/FLEFileChooserPlugin.m
evilBird: Done. I will create an issue to update the dart side with the suggested documentation, and to remove references to parameter keys that are not defined in the macOS plugin (title, message, canChooseFiles).

macos/library/FLEFileChooserPlugin.m
```diff
@@ -30,11 +30,15 @@
 static NSString *const kPlatformMethodNameKey = @"method";
 static NSString *const kPlatformMethodArgsKey = @"args";
 
-// Method argument keys.
+// Method argument keys for NSSavePanel.
+static NSString *const kOKButtonLabelKey = @"prompt";
```
stuartmorgan: My concern about this line being overly platform-specific was about the string value, which is the part that is a cross-platform API, not the variable name. The Dart and other desktop implementations should not have to call this argument "prompt" (which doesn't communicate much information about what it is) just because that's what the NSSavePanel calls the property.

That's the part I think should be called confirmButtonLabel or similar ("ok" is, again, macOS-specific)

macos/library/FLEFileChooserPlugin.m
```diff
@@ -30,11 +30,15 @@
...
+static NSString *const kInitialDirectoryKey = @"initialDirectory";
 static NSString *const kAllowedFileTypesKey = @"allowedFileTypes";
+static NSString *const kInitialFileName = @"nameFieldStringValue";
```
stuartmorgan: s/String/Initial/ (or initialFileName for better consistency with initialDirectory). We don't call the other strings "string", and the fact that it's the initial value is important information that the argument name should be capturing.

macos/library/FLEFileChooserPlugin.m
evilBird: That makes sense to me, thanks.

macos/library/FLEFileChooserPlugin.m
evilBird: Going with your suggestion of 'initialFileName' for consistency with 'initialDirectory'.

awdavies:
@stuartmorgan 

Updating now

awdavies:
@stuartmorgan PTAL.  The PR's are merged together

PR #53, opened 2018-05-10
[linux] Add file chooser additions.

awdavies:
Includes:
* Initial file name selection.
* Custom 'ok' button labeling.

This is to address #49


awdavies:
This is now merged into #49

PR #55, opened 2018-05-11
[linux] Update Makefiles for auto-building engine.

awdavies:
This also adds some missing info into the README, as well as removing
untrue documentation.


linux/Makefile
```diff
@@ -11,15 +11,57 @@
...
+
+# Runs gclient sync.
+$(FLUTTER_ENGINE_SYNC): $(FLUTTER_ENGINE_GCLIENT)
+	cd $(FLUTTER_ENGINE_ROOT) && $(GCLIENT) $(GCLIENT_ARGS)
+
+$(FLUTTER_ENGINE_GCLIENT): $(FLUTTER_ENGINE_GCLIENT_MANIFEST)
+	cp $(FLUTTER_ENGINE_GCLIENT_MANIFEST) $(FLUTTER_ENGINE_GCLIENT)
 
-.PHONY: $(TARGETS) $(SUBDIRS)
+.PHONY: $(TARGETS) $(SUBDIRS) clean clean_linux
```
awdavies: s/clean_linux/clean_local/

linux/engine.mk
```diff
@@ -0,0 +1,68 @@
...
+# build the flutter engine here:
+# https://github.com/flutter/engine/blob/master/CONTRIBUTING.md
+#
+# $(FLUTTER_ENGINE_ROOT) is the variable pointing to the git checkout of the
+# flutter engine. Here, the flutter engine path is relative to
+# `flutter-desktop-embedding/linux`
+#
+# This variable will be used when building `example_flutter/` to pass the
+# `--local-engine-src-path` flag to the flutter binary.
+FLUTTER_ENGINE_ROOT=../../engine
```
stuartmorgan: We should probably give this a less generic name (flutter-engine?) to avoid conflicts and/or confusion if this is part of a larger set of repositories someone is using, and then update the engine-cloning instructions accordingly.

linux/engine.mk
```diff
@@ -0,0 +1,68 @@
...
+# with the flags `--runtime-mode release`, then this variable will be set to
+# `host_release`.
+#
+# A simple way to figure out how to set this flag is to check the output of `gn`
+# and then set $(FLUTTER_ENGINE_BUILD) to be the name of the directory after `out/`
+#
+# Example output:
+#
+#```
+# $ ./src/flutter/gn --unoptimized
```
stuartmorgan: This is within the engine checkout, right? It would be worth making that explicit since the 'flutter' directory could be confusing otherwise.

linux/engine.mk
stuartmorgan: Shouldn't this and other variables that people might want to override be set with "?=" instead of "="?

linux/engine.mk
```diff
@@ -0,0 +1,68 @@
...
+# gn gen --check in out/host_debug_unopt
+# Done. Made 385 Targets from 168 files in 379ms
+#```
+#
+# From the above, you should then set this flag to `host_debug_unopt`.
+FLUTTER_ENGINE_BUILD=host_debug_unopt
+FLUTTER_ENGINE_SYNC=$(FLUTTER_ENGINE_ROOT)/src
```
stuartmorgan: Blank line above here? Everything after is derived from the above, rather than something potentially configurable, right?

linux/Makefile
```diff
@@ -11,15 +11,57 @@
...
+clean: clean_local
+	rm -f $(ENGINE_LIBRARY)
+	rm -f $(ENGINE_HEADER)
+	rm -f $(FLUTTER_ENGINE_GCLIENT) $(FLUTTER_ENGINE_ROOT_DIR)/.gclient_entries
+	rm -rf $(FLUTTER_ENGINE_SYNC)
+	rm -f $(ENGINE_LIBRARY)
+	rm -f $(ENGINE_HEADER)
+
+# Only cleans this folder (library and example)
+clean_local:
```
stuartmorgan: Is having 'clean' rm things outside this directory common? It seems potentially pretty surprising, so given that it's destructive I'd think we would want to err on the side of being conservative and have clean_all/clean rather than clean/clean_local.

linux/Makefile
```diff
@@ -11,15 +11,57 @@
...
+
+# Builds libflutter_engine.so
+$(FLUTTER_ENGINE_NINJA_CONFIG): $(FLUTTER_ENGINE_GN_TESTFILE)
+	$(FLUTTER_ENGINE_NINJA) $(FLUTTER_ENGINE_NINJA_ARGS)
+
+# Generates ninja files for building.
+$(FLUTTER_ENGINE_GN_TESTFILE): $(FLUTTER_ENGINE_SYNC)
+	$(FLUTTER_ENGINE_GN) $(FLUTTER_ENGINE_GN_ARGS)
+
+# Runs gclient sync.
```
stuartmorgan: gclient sync is a checkout command; having it run implicitly as part of a build step is not what I would expect. What's the reason for this?

linux/README.md
```diff
@@ -43,35 +44,15 @@ parent directory like so:
...
+`engine.version` file. Here is how you can check out the correct version of the
+engine when inside the `engine/` repo.
 
 ```
 $ git checkout $(cat ../flutter/bin/internal/engine.version)
 ```
 
-Then, follow the steps outlined on the flutter engine's
-[contributing](https://github.com/flutter/engine/blob/master/CONTRIBUTING.md)
-page so that you can build a copy of `libflutter_engine.so`.
```
stuartmorgan: You'll still need a modified version of this; in addition to the 'gclient sync' concern in my other comment, there are steps like sudo ./build/install-build-deps.sh that need to be run as part of initial setup.

linux/README.md
```diff
@@ -43,35 +44,15 @@ parent directory like so:
   └─ flutter-desktop-embedding/ (from https://github.com/google/flutter-desktop-embedding)
 ```
 
-Next you will need to build the flutter engine. First, after you've checked out
-the engine code, make sure to build the version appropriate for your checkout of
-Flutter. This can be done using the `engine.version` file. Here is how you can
-check out the correct version of the engine when inside the `engine/` repo.
+After you've checked out the engine code, make sure to build the version
```
stuartmorgan: s/build/sync to/ ?

linux/Makefile
awdavies: Ah, yeah I can see that being a bit of a misstep when trying to clean up things locally

linux/Makefile
awdavies: Only because it's required to get the correct version of the build tools used to compile the engine (want to make sure not to use the wrong binaries for ninja/gn (they have a specific setup that doesn't seem to generally work with a local version of gn).

Other than that it's just to ensure that the engine can be checked out and built.  This wouldn't happen regularly, e.g. once the user builds the engine once and doesn't change things locally as part of the linux embedder, the only thing that'd be built is the linux embedder.

linux/README.md
awdavies: Fixed.

linux/README.md
awdavies: Added!

linux/engine.mk
awdavies: Sounds like a good idea.  I'll update the README as well to reflect that.

awdavies: I wasn't aware that was possible!  Adding that in.

linux/engine.mk
awdavies: Fixed!

linux/engine.mk
awdavies: As of this comment, no.  I've updated it though so that there are more configurable targets.

linux/README.md
```diff
@@ -35,16 +35,17 @@ $ sudo apt-get install libglfw3-dev libepoxy-dev libjsoncpp-dev libgtk-3-dev \
 
 Also requires a checkout of the Flutter repo and Flutter Engine repo. For the
 most straightforward build make sure to set up all checkouts with the same
-parent directory like so:
+parent directory like so (note `flutter-engine` is not the default name when
+using `git checkout`. Make sure to check it out to the right directory):
```
awdavies: Fixed this to `git clone`

stuartmorgan:
Given that prebuilt Linux binaries are available now, let's drop this for now and instead add Linux support for getting the correct engine binary like we do on macOS.

We should keep this branch around though--there's still an open question as to whether the long-term solution should be building from source, and if we go that route later this patch will be useful.

awdavies:
Sounds good!  I'll close this for now, then.

vially:
> Given that prebuilt Linux binaries are available now

Where can we find these Linux binaries? 

/cc @stuartmorgan

Issue #13, opened 2018-02-28
MissingPluginException : the plugin works but an exception is throwed


rxlabz:
Hi, 
first : thanks for sharing this project ! it's really exciting :)
I started to play and tried to use plugins, but even if the call works, I get a MissingPluginException.

```dart
Future _colorPick() async {
  final channel = new MethodChannel('flutter/colorpanel', const JSONMethodCodec());
  final res = await channel.invokeMethod('ColorPanel.Show');
}
```

For instance when I use this method to open the colorPicker, the picker is opened, but I see an exception log.

```
[ERROR:topaz/lib/tonic/logging/dart_error.cc(16)] Unhandled exception:
MissingPluginException(No implementation found for method ColorPanel.Show on channel flutter/colorpanel)
#0      MethodChannel.invokeMethod (package:flutter/src/services/platform_channel.dart:153)
<asynchronous suspension>
#1      _MyHomePageState._colorPick (file:///Users/rxlabz/dev/labz/flutter/desktop/flutter_desktop2/example_flutter/lib/main.dart:81)
<asynchronous suspension>
```

Did I missed something ?

rxlabz:
the exception seems to be caused by the return of `nil` from the plugins methods... 🤔

stuartmorgan:
Issue #9 covers adding an example plugin, including the Dart side, to the example, to avoid people having to try to figure out the Dart side.

In this case, I think it's a plugin bug. The existing Dart code for interacting with this is using OptionalMethodChannel which means this never came up.

stuartmorgan:
Update on this: doing a quick fix for this is pretty easy, but in looking into it I filed issue #45 which ideally I'll do first.

It is definitely a plugin bug, which the current API makes much more likely than it should be.

stuartmorgan:
The plugins now return correct response objects, using the new API.

Issue #45, opened 2018-04-17
Provide a richer plugin API that handles responses


stuartmorgan:
Currently the fiddly details of constructing an appropriate response object to a platform channel call is left to each plugin. We should change the plugin API to provide functionality that abstracts away those details.

It would also be worth looking at any structural differences between the iOS/Android plugin channel API abstractions and the desktop plugin abstractions; ideally they should look as similar as possible so that writing plugins on desktop doesn't feel significantly different from writing them on mobile.

PR #56, opened 2018-05-17
Rework macOS plugin API

stuartmorgan:
Replaces the synchronous plugin response using raw data with an
asynchronous response using a much higher-level abstraction that makes
sending non-empty responses much easier.

The API changes are based very closely on the iOS Flutter plugin API,
so that it will be familiar to Flutter plugin developers, and make code
sharing easier. This includes using that APIs naming and terminology
whenever possible.

This is an incremental change, in that some aspects of the iOS API (most
notably message channel registration and method codecs) are still not
present in this API. A future change will further align the macOS API
with iOS, but this portion makes most of the changes that would
significantly change the structure of a plugin (allowing for rich, async
responses).

This addresses the macOS portion of issue #45.
Also fixes issue #13 on macOS.


macos/library/FLEChannels.h
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Initializes an FLEMethodError. If |details| is provided, it must be serializable to JSON.
+ */
+- (nonnull instancetype)initWithCode:(nonnull NSString *)code
+                             message:(nullable NSString *)message
+                             details:(nullable id)details NS_DESIGNATED_INITIALIZER;
+- (nonnull instancetype)init NS_UNAVAILABLE;
+
+/**
+ * The error code, as a string.
```
krisgiesing: What determines the namespace of error codes?

macos/library/FLEChannels.h
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * The name of the method being called.
+ */
+@property(readonly, nonatomic, nonnull) NSString *methodName;
+
+/**
+ * The arguments to the method being called, if any.
+ *
+ * This object must be serializable to JSON.
```
krisgiesing: You might want to note somewhere that NSJSONSerialization is what does the transformation.

macos/library/FLEChannels.h
```diff
@@ -0,0 +1,108 @@
...
+ */
+@property(readonly, nonatomic, nullable) id details;
+
+@end
+
+/**
+ * Returns the serialized JSON data that should be sent to the engine for the given
+ * FLEMethodResult argument.
+ *
+ * // TODO: Move this logic into method codecs.
```
krisgiesing: What's the benefit of moving this logic? I'm not disagreeing, just curious.

macos/library/FLEViewController.h
```diff
@@ -71,8 +71,12 @@
...
 
 /**
- * Sends a platform message on the given channel.
+ * Sends a platform message to the Flutter engine on the given channel.
+ *
+ * // TODO: Support responses.
  */
-- (BOOL)sendPlatformMessage:(nonnull NSData *)message onChannel:(nonnull NSString *)channel;
+- (void)invokeMethod:(nonnull NSString *)method
+           arguments:(nullable id)arguments
```
krisgiesing: nit: "withArguments:(nullable id)arguments" might be more idiomatic?

macos/library/FLEViewController.m
```diff
@@ -265,29 +285,43 @@ + (FlutterRendererConfig)createRenderConfigHeadless:(BOOL)headless {
...
-    id response = [plugin handlePlatformMessage:dictionary];
 
-    NSData *responseData = nil;
-    if (response != nil) {
-      responseData = [NSJSONSerialization dataWithJSONObject:response options:0 error:&error];
+  id messageObject = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
+  FLEMethodCall *methodCall =
+      messageObject ? [FLEMethodCall methodCallFromMessage:messageObject] : nil;
+
+  __block const FlutterPlatformMessageResponseHandle *responseHandle = message->response_handle;
```
krisgiesing: It might be worth a comment explaining that responseHandle isn't just a convenient intermediate variable but is captured in the block below. (Maybe this is obvious to people who write Objective-C more regularly though.)

macos/library/FLEChannels.h
stuartmorgan: I added a link to https://docs.flutter.io/flutter/services/JSONMessageCodec-class.html since that explicitly lists the supported types (and the goal is to move to having an FLEJSONMessageCodec, as well as versions of the others).

macos/library/FLEViewController.m
```diff
@@ -265,29 +285,43 @@ + (FlutterRendererConfig)createRenderConfigHeadless:(BOOL)headless {
...
-                                             responseData.length);
+  if (!methodCall) {
+    NSLog(@"Recieved invalid platform message: %@", messageObject);
+    resultHandler(FLEMethodNotImplemented);
+    return;
   }
+
+  id<FLEPlugin> plugin = _plugins[channel];
+  if (!plugin) {
+    resultHandler(FLEMethodNotImplemented);
```
awdavies: It seems like this would warrant a more specific error than FLEMethodNotImplemented?  Is this the same in the iOS implementation?

stuartmorgan: > Is this the same in the iOS implementation?

I believe so, yes:
https://github.com/flutter/engine/blob/58e84c8bf0b6304649340e6eaad988602961d7f6/shell/platform/darwin/ios/framework/Source/platform_message_router.mm#L38
An empty reply is what the higher-level code sends for FLEMethodNotImplemented.

PR #61, opened 2018-05-30
Rework Linux plugin API

stuartmorgan:
Updates Linux plugin API to the same level of mobile
consistency as the recent macOS rewrite.

Follows the Java repsonse objec pattern, since without
RTTI the ObjC approach of determining reponse type in
a single callback isn't viable.


stuartmorgan:
Text input wasn't working for me when I tested, but didn't work on trunk either so I'm not sure what's going on there. The other plugins work as expected.

linux/library/include/flutter_desktop_embedding/channels.h
```diff
@@ -0,0 +1,120 @@
...
+                     const std::string &error_message,
+                     const Json::Value &error_details) override;
+  void NotImplementedInternal() override;
+
+ private:
+  // Sends the given JSON response object to the engine.
+  void SendResponseJson(const Json::Value &response);
+
+  // Sends the given response data (which must either be nullptr or a serialized
+  // JSON response) to the engine.
```
awdavies: Please comment that this is because it is possible to send a success response that is still an empty string (and that is why this is a pointer instead of a const reference).

linux/library/src/channels.cc
```diff
@@ -0,0 +1,126 @@
...
+    const Json::Value &message) {
+  Json::Value method = message[kMessageMethodKey];
+  if (method.isNull()) {
+    return nullptr;
+  }
+  Json::Value arguments = message[kMessageArgumentsKey];
+  return std::make_unique<MethodCall>(method.asString(), arguments);
+}
+
+void MethodCall::set_arguments(const Json::Value &arguments) {
```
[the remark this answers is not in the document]
stuartmorgan: Nope, I just made name and arguments mutable by default. Now that you mention it there doesn't seem to be much of a point in them being mutable, so I'll remove both setters.

linux/library/include/flutter_desktop_embedding/channels.h
stuartmorgan: Done.

Issue #63, opened 2018-05-30
update_flutter_engine fails with a clean Flutter tree


stuartmorgan:
A fresh Flutter tree checkout doesn't have the dart binary it in, since it's fetched on demand. The wrapper script will need to run some minimal flutter command to fetch it (I'm not sure yet what the best option is).

PR #64, opened 2018-05-31
Ensure Dart is present in update_flutter_engine

stuartmorgan:
A clean Flutter doesn't contain Dart; check whether Dart is present
before trying to use it, and if not run 'flutter precache'.

Fixes issue #63

### DIFF
--- a/tools/update_flutter_engine
+++ b/tools/update_flutter_engine
@@ -14,4 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 base_dir="$(dirname "$0")"
-exec "$base_dir/../../flutter/bin/cache/dart-sdk/bin/dart" "$base_dir/dart_tools/bin/update_flutter_engine.dart" $@
+flutter_bin_dir="$base_dir/../../flutter/bin"
+dart_bin_dir="$flutter_bin_dir/cache/dart-sdk/bin"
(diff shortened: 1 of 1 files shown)


tools/update_flutter_engine
```diff
@@ -14,4 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 base_dir="$(dirname "$0")"
-exec "$base_dir/../../flutter/bin/cache/dart-sdk/bin/dart" "$base_dir/dart_tools/bin/update_flutter_engine.dart" $@
+flutter_bin_dir="$base_dir/../../flutter/bin"
+dart_bin_dir="$flutter_bin_dir/cache/dart-sdk/bin"
+# Ensure that the Dart SDK has been downloaded.
+if [[ ! -e $dart_bin_dir ]]; then
```
awdavies: wrap variable in quotes

stuartmorgan: Isn't one of the advantages of `[[ ]]` not needing to quote variables?

"Word splitting and filename expansion are not performed on the words between the [[ and ]]"
https://www.gnu.org/software/bash/manual/bashref.html#Conditional-Constructs

awdavies: Oh, woops!  I read that as single bracket.  Carry on

Issue #9, opened 2018-02-27
Add a plugin to the macOS example


stuartmorgan:
The example app should demonstrate registering a plugin with FLEViewController. A good candidate would be pulling the color picker out of the core set of plugins, making it optional, and using that as an example.

PR #69, opened 2018-06-13
Make the color panel a standalone plugin

stuartmorgan:
This extracts the color panel out of the embedder frameworks and into
a standalone plugin. As part of that, adds the Dart side of the plugin
to the respository.

Since this is the first plugin extracted, it serves as a template for
the plugin structure:
- Creates a new top-level plugins/ directory, with README.
- The plugin is a subdirectory containing:
  - The Dart code as a (local) package.
  - The macOS and Linux code, in macos/ and linux/ subdirectories
    (this code is moved from the framework directories).
    - Each has build files to create the native plugin library.
  - A common/ subdirectory for channel constants shared across
    platforms.
    - The duplicate strings have been extracted from the native
      code into a shared file.

Updates the example Flutter app to use the color plugin to show
how to import and use a plugin from the Dart side, as well as to
facilitate testing.

Since the Dart code is now part of the repository, this patch also
does some queued cleanup that required two-sided changes:
- Removes the unnecessary wrapping array from the color callback.
- Moves the normalization to 0-255 to the Dart side, so that it's
  not duplicated in the native code for each platform.
- Adds a callback for the case when the user closes the panel, so
  that the Dart-side understanding of whether the panel is showing
  stays accurate.

Fixes issues #9


plugins/color_panel/lib/color_panel.dart
```diff
@@ -0,0 +1,121 @@
...
+
+  /// Given a color component value from 0-1, converts it to an int
+  /// from 0-255.
+  int _colorComponentFloatToInt(num value) {
+    const colorMax = 255;
+    return (value * colorMax).toInt();
+  }
+
+  /// Mediates between the platform channel callback and the client callback.
+  Future<Null> _wrappedColorPanelCallback(MethodCall methodCall) {
```
awdavies: Does this need to be `Future<Null>`?  It looks like this function is synchronous.

plugins/color_panel/lib/color_panel.dart
```diff
@@ -0,0 +1,121 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+import 'dart:async';
+import 'dart:ui';
+
+import 'package:flutter/services.dart';
+
+const _kColorPanelChannel = 'flutter/colorpanel';
```
awdavies: These should probably have `String` declarations.

And _platformChannel down below should probably have a `const MethodChannel` declaration.

plugins/color_panel/lib/color_panel.dart
```diff
@@ -0,0 +1,121 @@
...
+      }
+    } on PlatformException catch (e) {
+      print('PLATFORM ERROR: ${e.message}');
+    }
+  }
+
+  /// Given a color component value from 0-1, converts it to an int
+  /// from 0-255.
+  int _colorComponentFloatToInt(num value) {
+    const colorMax = 255;
```
awdavies: const int?

plugins/color_panel/lib/color_panel.dart
awdavies: If we don't return an explicit `new Future` then I think we should add the `async` keyword, as that'll wrap the return value in a Future and prevent potential weird things from happening.

plugins/color_panel/lib/color_panel.dart
awdavies: Oh that's right, we talked about that.

plugins/color_panel/lib/color_panel.dart
stuartmorgan: Done

PR #71, opened 2018-06-15
Make the file chooser a standalone plugin

stuartmorgan:
This extracts the file chooser out of the embedder frameworks and into
a standalone plugin. As part of that, adds the Dart side of the plugin
to the respository. The overall structure duplicates that of the color
panel plugin.

Includes some adjustments to the color panel plugin due to issues that
weren't apparent until there was a second plugin:
- The channel constants need (distinct) namespaces to avoid runtime
  collision of identical constant names (e.g., kChannelName).
- To match those namespaces, the Linux plugin code now uses distinct
  namespaces for all of the code in each plugin instead of sharing the
  flutter_desktop_embedding namespace.
- Since the common headers are now C++, the macOS plugin implementations
  are now .mm instead of .m.
- That in turn surfaced latent errors due to missing Framework link
  commands (for the dependencies on AppKit and the embedding framework).

Since the Dart code is now part of the repository, this patch also
does some queued cleanup that required two-sided changes:
- Removes the callback, providing the response using the result of
  the initial method call now that the API for that is available.
  This simplifies the Dart code significantly.
- Removes the status code from the response, since it can be inferred
  by whether or not the array is null. Since there is now only one
  response element, it is returned directly instead of as part of
  an array.

Also restructures the Dart code into multiple files so that all
of the internal details of the channel controller class are not
mixed in with the public API. Some public API adjustments as well.

Lastly, updates the example Flutter app to have simple tests for
save and open: buttons that trigger each type of panel, and a
snackbar that shows the responses. No actual file reading/writing
happens since it's not relevant to exercising the panels.


stuartmorgan:
The Dart code has been significantly reworked (e.g., FileChooserConfigurationOptions is new, to reduce duplication in the show* commands), so that could use a careful pass.

Sorry about the macOS plugin implementation diff; between moving the constants out of the file, adding the namespace to all the keys, and eliminating a helper function that was no longer needed without the separate callback, it apparently was too much diff for the rename detector.

PR #79, opened 2018-06-26
Add input fields to example app

stuartmorgan:
This makes it easy to test basic text input support from the example
applications on each platform.

Includes two widgets since at some point in the future we can
hopefully support tabbing between fields, so having two will
be useful for testing.


stuartmorgan:
I'm not sure what to do with pubspec.lock. If we switch to tracking specific supported Flutter releases it'll be useful, but if we keep tracking head it's just noise (since every Dart version change unlocks dependencies). For now I'll keep it while I consider how to handle Flutter dependency versioning, so I went ahead and rolled it.

Issue #57, opened 2018-05-18
flutter-desktop do not display text in linux


pmatteo1998:
in the latest flutter revision (v0.4.5-pre.41) the Text() widget seems not render.
i've tested the feature both with "example_flutter" and an app write on-purpose to test the widget and not work anyway.

pmatteo1998:
i solved my self the issues , basically i put   roboto's ttf into the asset folder and edited font manifest.
i think that this must mentioned in the linux section to get at least the demo up & running correctly.

stuartmorgan:
@awdavies Can you take a look at this?

Before we add Roboto to the pubspec, it would be good to understand why this isn't an issue on macOS (even non-corp). Is Material choosing San Francisco as the font on macOS? If so, is that something we can depend on, or an accident of iOS detection?

awdavies:
@pmatteo1998 Can you link the versions of flutter and the engine so I can take a look?

Thanks.

pmatteo1998:
i don't know how to get the engine version but i'm using the version forked with my github account so i think you can get it from there

stuartmorgan:
Looks like the engine hash is 765b7d46b78fedbd94ed95c94b74e88c0a7fc758
And what's the hash of your Flutter tree?

Did you follow the instructions at https://github.com/google/flutter-desktop-embedding/tree/master/linux to ensure that you're using compatible versions of Flutter and the engine?

pmatteo1998:
yes,the command:
git checkout $(cat ../flutter/bin/internal/engine.version)
gives me:
HEAD is now at 06afdfe54... Ensure that application termination callbacks are serviced on the runner thread. (#5247)

stuartmorgan:
Okay, so you're not at the head of your fork of the engine tree then. Did you do the build with 06afdfe54..., or with 765b7d46b78fedbd94ed95c94b74e88c0a7fc758?

We still need the hash of your Flutter tree as well.

pmatteo1998:
Roboto is already present in android at system level so the app do not need to import it.
for ios flutter use a custom font which bundle strict into the asset output.

stuartmorgan:
I wasn't able to repro this, but expect that #82 fixed it. If you can still repro with latest trunk, please let me know.

PR #82, opened 2018-07-02
Set an explicit platform and font for the example

stuartmorgan:
This forces a target platform instead of relying on a host->target
mapping internal to Flutter (Fuchsia is used since it's not the
default for any desktop host platform, so puts everything on an
even footing), as well as setting the font to Roboto and bundling
Roboto.

This will make the example more consistent across platforms, and
should address Issue #57.


awdavies:
Great stuff!

PR #83, opened 2018-07-03
Add a .bat version of update_flutter_engine

stuartmorgan:
Allows easily running update_flutter_engine on Windows.

For now, since flutter_location isn't ported yet, uses a
hard-code directory for the flutter tree the way the other
platforms used to.


stuartmorgan:
@clarkezone FYI. Also, any comments welcome since this is my first attempt at a .bat file :)

stuartmorgan:
I'll go ahead and land this; if there are .bat best practices it's not following it can be improved in follow-ups.

PR #84, opened 2018-07-04
make edits in README.md

melongbob:
- "... contains basic embeders for Flutter on desktop platforms, ... for building desktop applications that embed Flutter."
"desktop" and "embed flutter" was removed from the supporting clause to reduce redundancy. It is mentioned in the first part of the sentence that "native application" is on desktop, and it embeds Flutter

- "Currently macOS and Linux are supported, and the goal is to support Windows..."
"the goal is to..." clause is not equivalent to "macOS and Linux are supported". The sentence is broken down into two sentences to separate two independent ideas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,8 @@
 
 This repository contains code that implements basic embedders for
 [Flutter](https://github.com/flutter/flutter) on desktop platforms, as starting
-points for building native desktop applications that embed Flutter.
-Currently macOS and Linux are supported, and the goal is to support Windows
+points for building native applications.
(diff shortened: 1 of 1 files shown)


stuartmorgan:
> to reduce redundancy

It also reduces clarity, since Flutter is primarily known as a project for building native mobile applications. A desktop embedder could potentially be used for developing mobile applications, for example by replacing the simulator (and in fact people have expressed interest in exactly that).

Being clear about the purpose of the project is more important to me that having the sentence be a few words shorter.

> to separate two independent ideas

These are not independent ideas.

PR #85, opened 2018-07-05
Make the project compile under Ubuntu 16.04 or lower

badpx:
When the project compiling under Ubuntu 16.04, it prompts that some symbol definitions are missing, because Ubuntu 16.04 can only install the older version of **libglfw3-dev(v3.1)** and **libgtk-3-dev(v3.18)** via `apt-get`. Make the project can be compiled by the following modifications:

1. Add the `GLFW_TRUE` macro when it's undefined.
2. Replace `GtkFileChooserNative` with `GtkFileChooserDialog` to support lower GTK versions.


badpx:
I signed it!

stuartmorgan:
In general I'm not sure we want to commit to supporting compilation on older OS versions; this change is fairly minor (although already reduces flexibility somewhat, based on the documentation, which is not ideal) but I'm not sure how many limits supporting 16 would place on the full GTK implementation.

Is there a compelling need to support older versions?

awdavies:
I wasn't able to find the branch on my fork, so I'd have to go digging around on my machine.  I don't see a major issue with this as long as behavior is the same on Linux systems (I'm not sure if I remember whether or not this would raise an error in the event that no parent window is specified.  That was the main reason I implemented this using the native file chooser dialogue).

plugins/file_chooser/linux/src/file_chooser_plugin.cc
```diff
@@ -159,7 +161,7 @@ void FileChooserPlugin::HandleMethodCall(const MethodCall &method_call,
     result->NotImplemented();
     return;
   }
-  chooser_res = gtk_native_dialog_run(GTK_NATIVE_DIALOG(chooser));
+  chooser_res = gtk_dialog_run(GTK_DIALOG(chooser));
```
awdavies: As a result, I believe on line 181 we'll need to run `gtk_widget_destroy` instead of `g_object_unref`. This appears to be the way the example works, anyway.  Can't remember the specifics around destroying the widget vs un-reffing the object.

[Regular dialog](https://developer.gnome.org/gtk3/stable/GtkFileChooserDialog.html) vs [Native](https://developer.gnome.org/gtk3/stable/gtk3-GtkFileChooserNative.html)

plugins/file_chooser/linux/src/file_chooser_plugin.cc
```diff
@@ -86,19 +86,21 @@ static void ProcessAttributes(const Json::Value &method_args,
 // string, then this returns a file saver dialog.
 //
 // If the method is not recognized as one of those above, will return a nullptr.
-static GtkFileChooserNative *CreateFileChooserFromMethod(
+static GtkFileChooserDialog *CreateFileChooserFromMethod(
     const std::string &method, const std::string &ok_button) {
-  GtkFileChooserNative *chooser = nullptr;
+  GtkFileChooserDialog *chooser = nullptr;
```
awdavies: nit: Perhaps we can change this to `GtkWidget*` and get rid of the casting, then return `GTK_DIALOG(chooser)` at the end of the function?  I'm okay with this implementation either way.

awdavies:
@badpx Pingin'

Not sure if the review sent out an email or not.

plugins/file_chooser/linux/src/file_chooser_plugin.cc
badpx: Sorry, it‘s my fault, it must be changed to `gtk_widget_destroy`.

plugins/file_chooser/linux/src/file_chooser_plugin.cc
badpx: This is really a good suggestion, I will modify this function as you said.

stuartmorgan:
As far as git is concerned master is just a branch like any other. You can keep using this, and then clean up your fork after the PR is closed.

Could you answer the questions above that aren't about specific lines of code:

- What's the reason for supporting a release other than the latest as a compilation platform?
- Have you verified that the file chooser works correctly in the example app?

badpx:
Apologize for not answering that two questions:
1. I just think 16.04 might be more stable because it's the last LTS version before 18.04(and still in support), hardware driver support is also more complete at least for now(eg CUDA). Maybe there are still a lot of computers running it. If this is the case, it's easier to modify the code of this repository than to recompile the new version of GTK from source.
2. I verified on 16.04 that the file chooser works correctly, but it has not been verified on a later versions because I have not installed them.

The answer to the first question is purely personal, 16.04 is indeed a very old version, it might be better to give it up. Thank you very much!

stuartmorgan:
Fair enough, the changes here are relatively minor so I'm fine with going forward with the PR, with the caveat that I don't guarantee that we'll keep 16.x compatibility for the full Linux implementation.

@awdavies can you give this a spin on 18.x when you do the next review pass to make sure everything still looks good there?

awdavies:
Looks good, and it builds.  Just haven't verified the plugin behavior is the same yet

badpx:
If there is anything I can do, please let me know, thank you.

stuartmorgan:
The plugin behavior still looks good in the sample app. If you can resolve the new conflicts (sorry about that), this is good to land!

stuartmorgan:
@awdavies Can you mark your review as approved for bookkeeping purposes?

badpx:
@stuartmorgan Thanks for your verification! Now all conflicts have been resolved.
@awdavies Thank your for the patient review!

PR #100, opened 2018-07-10
Set debugDefaultTargetPlatformOverride

stuartmorgan:
Flutter no longer maps desktop host platforms to mobile target
platforms, and instead throws 'Unknown platform' exceptions.
This adds a debugDefaultTargetPlatformOverride to bypass the
detection logic so that the example runs on desktop.

Also updates the main README to include a section on likely pitfalls
when attempting to run a stock Flutter application on desktop (so far,
this issue and font handling), along with suggested solutions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,71 @@ In addition, there is:
   by the build systems, but in the future developer utilities providing
   some functionality similar to the `flutter` tool may be added.
 
+## Flutter Application Requirements
+
+Because desktop platforms are not supported Flutter targets, existing Flutter
+applications are likely to require slight modifications to run.
+
+### Target Platform Override
+
+Most applications will need to override the target platform for the application
+to one of the supported values in order to avoid 'Unknown platform' exceptions.
+This should be done as early as possible.
+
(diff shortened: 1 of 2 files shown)


stuartmorgan:
Prep for https://github.com/flutter/flutter/pull/19187 which sounds like it's landing shortly.

awdavies:
Should we specify that we need a minimum version for Flutter?  Or is the assumption still that the user should always have their dependent flutter version as up to date as possible?

stuartmorgan:
> Or is the assumption still that the user should always have their dependent flutter version as up to date as possible?

For now I've been operating on that assumption, but I've been thinking about adding a general note to the main README to that effect. I can do that in a follow-up.

This specifically should be pretty harmless in versions that don't have the change though, since the variable exists already, but has previously been ignored in release builds. For people without the latest Flutter it should just drop back to the state where the example app uses iOS on macOS and Android on Linux, which isn't that big a deal.

awdavies:
Sounds good!

PR #106, opened 2018-07-13
Add User Specific Files For Debugging

CallumIddon:
This will include the debugging configuration required for VS to correctly run the embedder from inside VS. This includes the configuration that the working directory needs to be the solution directory not the project directory.


stuartmorgan:
I don't have access to my Windows machine at the moment to test; does
https://stackoverflow.com/questions/38295289/how-to-store-visual-studios-debug-configuration-in-source-control
still work in VS 2017 so that we wouldn't need to check in the user-specific file?

CallumIddon:
Works for me if you could test it just in case that would be great.

b-strauss:
Works great on my machine with Visual Studio Community!

gedw99:
Ah that's great to hear :)

gedw99:
I presume by VS you don't mean VS Code ? Just checking

CallumIddon:
@stuartmorgan should be good to land. And yes this is Visual Studio not Visual Studio Code.

stuartmorgan:
Great, will land now since it's confirmed working (I wasn't going to be able to test until Monday).

PR #110, opened 2018-07-17
Extract codecs from plugin API

stuartmorgan:
The splitting of channel.* into separate files for each class, and the fact that plugin_handler was changed enough that it wasn't recognized as a move, makes the Linux change look larger than it actually is (as does the define guard and include path churn). If that gets in the way of reviewing let me know and I can retroactively create a prequel that does all the source rearranging and include path cleanup without any of the code changes.

### DIFF
--- a/linux/library/src/internal/engine_method_result.h
+++ b/linux/library/src/internal/engine_method_result.h
@@ -39,6 +39,7 @@ class EngineMethodResult : public MethodResult {
   ~EngineMethodResult();
 
  protected:
+  // MethodResult:
   void SuccessInternal(const void *result) override;
   void ErrorInternal(const std::string &error_code,
                      const std::string &error_message,
--- a/linux/library/src/internal/engine_method_result.h
+++ b/linux/library/src/internal/engine_method_result.h
@@ -1,0 +1,61 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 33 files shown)


stuartmorgan:
Rebased onto the Linux cleanup commits; the Linux change should be less painful to review now.

linux/library/src/internal/engine_method_result.h
```diff
@@ -0,0 +1,61 @@
...
+  //
+  // If the codec is null, only NotImplemented() may be called.
+  EngineMethodResult(
+      FlutterEngine engine,
+      const FlutterPlatformMessageResponseHandle *response_handle,
+      const MethodCodec *codec);
+  ~EngineMethodResult();
+
+ protected:
+  void SuccessInternal(const void *result) override;
```
awdavies: Add a comment for the class which is being overridden.

macos/library/FLECodecs.h
```diff
@@ -0,0 +1,61 @@
...
+- (nullable FLEMethodCall*)decodeMethodCall:(nonnull NSData*)methodData;
+
+/**
+ * Returns a binary encoding of |result|. Returns nil if |result| is not a type supported by the
+ * codec.
+ */
+- (nullable NSData*)encodeSuccessEnvelope:(nullable id)result;
+
+/**
+ * Returns a binary encoding of |error|. Returns nil if the |details| parameter of |erorr| is not
```
awdavies: s/erorr/error/

PR #122, opened 2018-08-21
macOS library code cleanup

stuartmorgan:
Miscellaneous cleanup for the macOS embedding framework:
- Make FLEViewController's conformance to FLEReshapeListener public
  since that would be necessary for a non-FLEView view implementation.
- Update files in Xcode so that only public headers are in the Public
  group
- Reorganize FLEViewController's methods so that the #pragma groupings
  are accurate, and so that the definition orders match the declaration
  orders.
- Declare all private methods expplicitly in the class extension, and
  move all declaration comments there, to make the internal API surface
  clearer.
- Remove nullability annotations from the implementations, since they
  are intended for interfaces and mismatches can be problematic.
- Rename "theEvent" to "event" to match normal variable naming practice.

### DIFF
--- a/macos/library/FLEViewController.m
+++ b/macos/library/FLEViewController.m
@@ -34,23 +34,13 @@ static const int kDefaultWindowFramebuffer = 0;
 /**
  * Private interface declaration for FLEViewController.
  */
-@interface FLEViewController () <FLEReshapeListener> {
+@interface FLEViewController () {
   NSOpenGLContext *_resourceContext;
 }
 
 /**
- * Responds to system messages sent to this controller from the Flutter engine.
- * Invoked from a static method provided in the engine configuration, thus must be
- * explicitly declared.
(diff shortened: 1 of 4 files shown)


stuartmorgan:
While the diff looks large, it's almost all just reordering of methods.

macos/library/FLEViewController.m
```diff
@@ -34,33 +34,81 @@
...
+@property NSMutableOrderedSet<NSResponder *> *additionalKeyResponders;
+
+/**
+ * Initializes plugins used by this view controller.
+ * Called from init.
+ */
+- (void)initPlugins;
+
+/**
+ * Internal implementation of addPlugin:. Separated since it is called during init, so must not
```
awdavies: superfluous period?

stuartmorgan: Nope, "addPlugin:" is the name of the method.

PR #125, opened 2018-08-28
Add a wrapper script for 'flutter build bundle'

stuartmorgan:
This consolidates the logic for creating the flutter_assets directory in
a central location, rather than it being embedded into each
platform-specific project. Benefits:
- Easier to update if the command changes, as happend with the change
  from flx to bundle. Clients can use this script and be insulated from
  any such changes in the future.
- Easier to understand the details of what's run without having to find
  it in the native projects (e.g., in Xcode, where it's non-trivial to
  find especially for people not familiar with Xcode's UI).
- Easier to change to a local engine build for testing, due to including
  the necessary flags as commended-out variables. (In the future, if
  it's sufficient useful, the scripts could be extended to take a config
  file as with Flutter location.)

### DIFF
--- a/linux/example/Makefile
+++ b/linux/example/Makefile
@@ -16,7 +16,8 @@
 FLUTTER_EMBEDDER_LIB_DIR=$(CURDIR)/../library
 FLUTTER_APP_DIR=$(CURDIR)/../../example_flutter
 PLUGINS_DIR=$(CURDIR)/../../plugins
-FLUTTER_DIR=$(shell ../../tools/flutter_location)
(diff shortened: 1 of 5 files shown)


stuartmorgan:
I've been doing work with a local engine, and found that changing the Xcode project was annoying. I thought this would be a good opportunity to make the process of using a local engine clearer.

James, since you've been dealing with this I'd be curious to know if you think this is more straightforward. (It could also use docs on the process, but I want to break out a more technical doc from the top-level README before adding more documentation).

PR #127, opened 2018-09-24
[macos] Simplify view controller initialization

stuartmorgan:
Moves internal plugin registration to just before the engine starts
rather than during init, to avoid the complexity of having plugin
registration potentially happening during init.

Also moves common ivar initialization from an ObjC method to a static
function, which is a safer pattern as it isn't subject to naming
collisions in subclasses.


krisgiesing:
LGTM!

PR #128, opened 2018-09-24
[macos] Move FLEMethodCall and FLEMethodError

stuartmorgan:
Moves FLEMethodCall and FLEMethodError out of Channels.h. This better
aligns with the Flutter iOS structure, although this deviates in that on
iOS these classes live in the Codecs.* files, while this keeps them
separate to reduce the number of classes being declared in a single
header.

Preparation for the remaining plugin refactoring (which adds channel
classes to the Channels.* files).

### DIFF
--- a/macos/library/FLEChannels.h
+++ b/macos/library/FLEChannels.h
@@ -15,38 +15,8 @@
 #import <Foundation/Foundation.h>
 
 /**
- * An object encapsulating a method call from Flutter.
- */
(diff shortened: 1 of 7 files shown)


stuartmorgan:
I'm not sure why the new files aren't being shown as slightly modified copies; the code itself is unchanged.

krisgiesing:
LGTM!

PR #133, opened 2018-10-06
[macos] Add MethodChannel

stuartmorgan:
Adds the MethodChannel class, and the accompanying BinaryMessegener
protocol, for parity with the existing iOS Flutter plugin/channel
structure.

This lays the foundation for a future breaking change that makes plugin
creation and registration match the iOS structure, but for now makes the
method channel an internal implementation detail of the existing plugin
messaging design to preserve compatibiility.

### DIFF
--- a/macos/library/FLEViewController.m
+++ b/macos/library/FLEViewController.m
@@ -36,12 +36,10 @@ static const int kDefaultWindowFramebuffer = 0;
  */
 @interface FLEViewController () {
   NSOpenGLContext *_resourceContext;
-}
 
-/**
- * The registered FLEPlugin instances, keyed by the channel they are registered on.
- */
-@property NSMutableDictionary<NSString *, id<FLEPlugin>> *plugins;
+  // A mapping of channel names to the registered handlers for those channels.
+  NSMutableDictionary<NSString *, FLEBinaryMessageHandler> *_messageHandlers;
+}
 
 /**
  * A list of additional responders to keyboard events. Keybord events are forwarded to all of them.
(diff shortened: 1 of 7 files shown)


stuartmorgan:
Originally I was going to do the Linux version of this at the same time, but since this step doesn't break anything at the API layer I decided there's no reason not to do it piecemeal as I have time to implement it for each language.

macos/library/FLEViewController.m
```diff
@@ -227,28 +220,45 @@ - (BOOL)launchEngineWithMainPath:(NSURL *)main
...
   return YES;
 }
 
 - (void)invokeMethod:(NSString *)method
            arguments:(id)arguments
-           onChannel:(NSString *)channel
+           onChannel:(NSString *)channelName
            withCodec:(id<FLEMethodCodec>)codec {
-  FLEMethodCall *methodCall = [[FLEMethodCall alloc] initWithMethodName:method arguments:arguments];
-  NSData *messageData = [codec encodeMethodCall:methodCall];
```
krisgiesing: I know it's not the point of this PR, but where is this object released?

krisgiesing:
LGTM, although I have a lifecycle question (the answer to which is probably not affected by this PR).

macos/library/FLEViewController.m
stuartmorgan: `messageData`? Nothing keeps a reference to it, so it'll be released when it goes out of scope. The engine doesn't require that the underlying bytes stay valid beyond the call to FlutterEngineSendPlatformMessage.

PR #137, opened 2018-10-12
Restructure the repository to make platform secondary

stuartmorgan:
This change flips the structure of the repository so that instead of top-level platform directories each containing `library` and `example` directories, those two become top-level and the platform divisions are underneath them.

This has several advantages:

- There's a much clearer path to sharing code between platforms in the library (e.g., plugin code on Windows and Linux should be largely the same)
- It's better suited to a cross-platform build system (i.e., GN).
- It better groups the example code that's not actually needed by a client, especially the previously-top-level Flutter example application.


krisgiesing:
I guess there are no logic changes here? (The diff is very large but I think it's all due to file movements.) Assuming that's true, LGTM.

stuartmorgan:
This will be a breaking change for projects including it and linking to it, so I'll send out an email on the list when it lands.

Most of the change is just file moves. Substantive changes:

- The READMEs were restructured and somewhat rewritten for the new layout.
- The Linux top-level Makefile is gone, in favor of the example app's Makefile building the necessary sub-pieces.
- There's a new VS solution for just the library, in addition to the existing one that contains the example project and library project.
- Some Windows build scripts were made relative to themselves rather than the current working directory, and some project settings were changed from SolutionDir to ProjectDir, to avoid breaking when used from different solutions.

stuartmorgan:
I'm going to go ahead and land this since it will conflict with basically almost any other patch. The most likely things that would be caught in a review are README changes and those are easy to clean up after the fact.

krisgiesing:
Sorry, forgot to re-LGTM after your response. Thanks for updating the documentation!

Issue #132, opened 2018-10-02
Build macOS Example Embedder failed in Xcode 10


chenyuantao:
Showing Recent Messages
Ld /Users/Tao/Library/Developer/Xcode/DerivedData/Example_Embedder-cyebqnokskymrzftfrnhdolxhaqu/Build/Products/Debug/Example\ Embedder.app/Contents/MacOS/Example\ Embedder normal x86_64 (in target: Example Embedder)
    cd /Users/Tao/development/flutter-desktop-embedding/macos/example
    export MACOSX_DEPLOYMENT_TARGET=10.13
[...]

PR #148, opened 2018-11-05
Add plugin frameworks to macOS example link step

stuartmorgan:
Speculative fix for issue #132


stuartmorgan:
I hit #132 myself, and noticed that the frameworks weren't actually listed as link dependencies (which moves this into "how did this work before" territory). I can't figure out how to recreate the issue again, so I'm not certain that this fixes it, but it did seem to resolve it when I did have the build failure.

krisgiesing:
LGTM

krisgiesing:
The libraries are listed in the project config as "embedded binaries", maybe that implicitly declared a dependency in Xcode 9.x but not in 10?

Issue #114, opened 2018-07-27
Use a cross-platform build system


DavBfr:
Using CMake would make it easier to build multiplatform projects and use xcode / vstudio if needed only.

stuartmorgan:
This is related to #48, although that discussion is more about a higher-level tooling abstraction.

Although CMake would be a possibility, I'm actually already strongly considering moving to gn (I have a mostly complete implementation for Linux already in a local branch). It's mostly a question of making sure that the gn dependency doesn't become a significant pain point.

mrmcq2u:
meson might be nice compromise :)

MikeMitterer:
Please! Use a standard build system like Gradle or cmake not an exotic one that nobody knows except a few needs from a special community.

stuartmorgan:
> not an exotic one

The Flutter engine, which is where the iOS and Android platform implementations that this project parallels live, uses GN, so it's not exotic in this context.

Also, keep in mind that, as noted in the library README, "a more Flutter-like model of using an SDK containing pre-compiled binaries is likely to be supported in the future". I.e., most clients of the library will never know nor care what build system it uses.

sirgallifrey:
> I'd encourage anyone on Linux who is interested in this issue to try out:
> https://github.com/google/flutter-desktop-embedding/blob/master/library/GN.md
> and let me know if there are any pain points. (Since I originally looked at GN, they've added prebuilt binaries, which makes things much simpler.)

I just followed the GN docs and successfully built both the lib and the example. No pain points, but I hadn't any pain points with make either.

Also for the exotic part, I had never heard of GN or Ninja, just followed the instructions which are super simple.

charafau:
> Please! Use a standard build system like Gradle or cmake not an exotic one that nobody knows except a few nerds from a special community.

I know I don't have much power here, but as developer I prefer to have gn and ninja instead of gradle, which requires java to run.

stuartmorgan:
I forgot to tag this issue in #262, but with Windows and Linux switched to GN, and macOS on GN in flutter/engine (the copy here will be removed shortly, so won't be switched), this is essentially done.

Per my comment above the only thing not on GN is macOS plugins, but until there's a stabilized plan for plugin management on the desktop platforms it doesn't make sense to switch them to GN (notably, if they are distributed the way iOS plugins are currently distributed, Xcode builds would be necessary).

PR #149, opened 2018-11-05
Initial Linux GN build

stuartmorgan:
Adds an experimental GN-based build as an alternative to the Make-based
build on Linux.

The example application does not use GN for its own build since it is
intended to serve as an example of using the embedding library, so it
continues to use the more common Make as clients are unlikely to be
using GN.

First step toward possible approach to #114

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
+out/
 .DS_Store
--- a/.gn
+++ b/.gn
@@ -1,0 +1,1 @@
+buildconfig = "//build/BUILDCONFIG.gn"
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1,0 +1,22 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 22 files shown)


krisgiesing:
Sorry I haven't had time to take a look at this. I may not be the best reviewer anyway since I don't know a lot about GN. I'll try to find a reviewer with more knowledge.

stuartmorgan:
FWIW, I'm fine a reviewer that's not a GN expert; I can always look for an expert GN review later if this does turn out to be a viable replacement for the builds across all three platforms.

krisgiesing:
RSLGTM. Just so I understand, the intent here is for those building the library to use GN, but those consuming the library may use other systems (e.g. Make on Linux, Xcode on Mac)? So we are looking ahead to a system where we might publish prebuilt libraries to clients rather than having them build from source?

stuartmorgan:
> So we are looking ahead to a system where we might publish prebuilt libraries to clients rather than having them build from source?

Exactly. Since that's the model used by Flutter, it seems like the natural direction for this project to go in longer term so that the workflows match the expectations of existing Flutter developers.

krisgiesing:
> Exactly. Since that's the model used by Flutter, it seems like the natural direction for this project to go in longer term so that the workflows match the expectations of existing Flutter developers.

Sounds good to me. Thanks for clarifying!

Issue #102, opened 2018-07-12
Make the plugin API structure match the existing Flutter plugin API


stuartmorgan:
This is essentially a continuation of issue #45. I'd like the whole method channel/plugin API on desktop match the existing Flutter API as much as possible. The current implementation grew out of a very minimal initial macOS version; the changes in #45 started to bring it in line with the iOS and Android implementations, but there's more to do.

stuartmorgan:
With those last two patches, I'm going to call this complete. There are still missing pieces to be added later, but there's no longer a structural difference between how plugins are written for desktop and how they are written for mobile. The APIs that are there should now seem familiar to people who have written Flutter plugins for mobile in the past.

PR #162, opened 2018-12-05
 [macos] Add BasicMessageChannel

stuartmorgan:
Further API alignment (issue #102). Adds support for communicating with
a BasicMessageChannel on the Flutter side.

Also splits apart the files related to codecs and channels to be one class
per file. They were originally combined to match the Flutter iOS file structure,
but that has proven to hinder project navigation more than it helps with
cross-referencing.

Moved classes have been cleaned up to be more consistent about using
nullability declarations only on interfaces, not implementations.

### DIFF
--- a/library/macos/FLEMethodChannel.h
+++ b/library/macos/FLEMethodChannel.h
@@ -15,59 +15,7 @@
 #import <Foundation/Foundation.h>
 
 #import "FLEBinaryMessenger.h"
-#import "FLECodecs.h"
-#import "FLEMethods.h"
-
-/**
- * A message response callback. Used for sending a message's reply back to the Flutter engine.
- * The reply must be serializable by the codec used to encode the message.
- */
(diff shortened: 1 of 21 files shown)


library/macos/FLEMethodChannel.h
```diff
@@ -69,14 +69,14 @@ extern NSString const *_Nonnull FLEMethodNotImplemented;
  *
  * The arguments, if any, must be encodable using this channel's codec.
  */
-- (void)invokeMethod:(nonnull NSString *)method arguments:(id _Nullable)arguments;
```
franciscojma86: Should the qualifiers in FLEBasicMessageChannel.h change as well?

stuartmorgan: The nullability annotations on FLEBasicMessageChannel are using this style. The typedefs (as with those at the top of this file) can't, because this form isn't usable in all language constructions.

Issue #164, opened 2018-12-06
IDE Support


amrboxit4me:
I wonder when it will have IDE Plugin for building the project to target a specific desktop platform?

stuartmorgan:
To make sure I'm understanding correctly, you're referring to being able to open a Flutter project in something like VS Code or IntelliJ and doing a build-and-run to launch the desktop version?

Ideally the approach for that will be to make the `flutter` tool extensible in a way that would allow existing IDE support to work with desktop, rather than writing new plugins for each IDE. I don't have a specific timeline for that, but it's certainly high on the priority list.

amrboxit4me:
@stuartmorgan Yes that's what I mean exactly 
You know not everyone is advanced in building system and terminal and these things

stuartmorgan:
There's nothing else tracking this yet, so it should definitely stay open.

gabrc52:
I'm using a simple bash script and the `terminal-command-keys` plugin in the meantime: https://gist.github.com/gabrc52/ec0ea109322c816e7a81f0f58afabd24

alexissantos:
+1 To this. Lookin' forward to it.

stuartmorgan:
Moved to https://github.com/flutter/flutter/issues/30707 (and the related 30703-6)

Issue #14, opened 2018-03-01
Hot reload support?


szotp:
Hi, I'm interested in adding support for hot reload and debugging, but I don't know enough about flutter/dart ecosystem. Would implementing https://github.com/flutter/flutter/issues/5043 solve the problem  or is it more complicated?

Something like `flutter connect http://127.0.0.1:49278/` to completely bypass device detection, etc.

szotp:
Did some research on this. Not sure if it will be useful to anyone

* I added fake device and hardcoded executable path and Uri on both sides. `flutter run` would then launch the .app  and correctly perform the 'r' reload. 'R' reload did not work.

* full reload  did not work because it is not handled in the embedder: https://github.com/flutter/engine/blob/master/shell/platform/embedder/platform_view_embedder.cc#L53. I could make it work by copying some code from the ios platform view. This made flutter run + R work.

* VSCode extension uses daemon, which somehow performs hot reload and restart differently than the run command. Nor debugging nor reloading would work. Something about invalid UTF8 would appear when trying to reload.

stuartmorgan:
I haven't looked at reload support yet, so I don't have any extra insight into what all is necessary. I think we'll need to discuss with Flutter developers how reloading should work in the embedding model though (e.g., it might make sense for it to be exposed as an embedding API). At a high level, I don't think we want people to have to launch applications with `flutter run` to get tooling support.

szotp:
Thanks for the answer. I got it to work somehow. Perhaps by switching to flutter's master branch. For anyone interested these are the steps required:

1. Create hardcoded desktop device. I used port 5300. 
https://github.com/szotp/flutter/commit/5c06b5dd2469249cbcd3e55957458d53d1714375
2. Supply these arguments in the client:
`[CommandLine.arguments[0], "--observatory-port=50300", "--start-paused"]`
3. Launch the app from Xcode.
4. Open the project_example in Visual Studio Code. Make sure it works on iOS (plist needed).
5. Press F5. Debugging and hot reload should work. Restart reload requires some changes in the engine.

I think I could improve the fork and prepare it for merge:
1. Add flag in the config to enable the desktop device so only people interested in it can see it.
2. Execute some script in the flutter, perhaps $(PROJECT_DIR)/launch.{sh/bat} that will compile and launch the app, so we don't have to do it manually via Xcode. Script would also return an Uri to connect.

stuartmorgan:
[...]

PR #168, opened 2018-12-06
Add initial debugging instructions

stuartmorgan:
Provides basic instructions for debugging the Flutter side of a desktop
application, since the normal approaches for Flutter debugging require
some modifications on desktop.

Related to issue #14 and issue #164, although doesn't fully address
either.

### DIFF
--- a/Debugging.md
+++ b/Debugging.md
@@ -39,7 +39,9 @@ There are two options:
 ### Command Line
 
 The `flutter attach` command can connect to a desktop Flutter application
-and provide the same interactive command line that `flutter run` would provide:
+and provide the same interactive command line that `flutter run` would provide.
+In the directory of the Flutter portion of your application (e.g.,
+`/example/flutter_app`) run:
 
 ```
 $ flutter attach --device-id=flutter-tester --debug-port=49494
@@ -68,7 +70,7 @@ like the following, substituting your Observatory port:
 You will likely want to hard-code your observatory port, otherwise you will
 need to change `launch.json` every time you relaunch the app.
 
-In addition to the Flutter debug commands, source-level debugging with pausee,
+In addition to the Flutter debug commands, source-level debugging with pause,
 continue, variable inspection, etc. should all work.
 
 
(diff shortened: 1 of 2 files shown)


Debugging.md
```diff
@@ -0,0 +1,78 @@
...
+
+Open the Flutter portion of your application (e.g., `/example/flutter_app`).
+Add a [launch
+configuration](https://code.visualstudio.com/docs/editor/debugging#_launch-configurations)
+like the following, substituting your Observatory port:
+
+```
+  {
+    "name": "Flutter Desktop Attach",
+    "request": "attach",
```
franciscojma86: This didn't attach for me until I added "program" : "example/flutter_app/lib/main.dart". Is this the case for you as well

Debugging.md
```diff
@@ -0,0 +1,78 @@
...
+    "deviceId": "flutter-tester",
+    "observatoryUri": "http://127.0.0.1:49494/",
+    "type": "dart"
+  }
+```
+
+You will likely want to hard-code your observatory port, otherwise you will
+need to change `launch.json` every time you relaunch the app.
+
+In addition to the Flutter debug commands, source-level debugging with pausee,
```
franciscojma86: pause*

Debugging.md
```diff
@@ -0,0 +1,78 @@
...
+   engine. If you are using `example/`, or code based on it, look for the
+   line that adds `--dart-non-checked-mode` for an example of adding arguments.
+   (Be sure not to add the `observatory-port` argument inside the `#if`,
+   however.)
+
+## Debugging
+
+### Command Line
+
+The `flutter attach` command can connect to a desktop Flutter application
```
franciscojma86: Not sure if this is obvious, but might be worth clarifying that this command should be called from a directory that contains 'lib/main.dart'. I was getting an error when trying from another directory.

Debugging.md
franciscojma86: Yeah, I was still in root since I was already working there. Thanks!

Issue #173, opened 2018-12-17
[macOS] Add FLEEventChannel


cloudwebrtc:
Adds support for communicating with
a EventChannel on the Flutter side.

PR #174, opened 2018-12-17
[macOS] Add FLEEventChannel

cloudwebrtc:
Adds support for communicating with
a EventChannel on the Flutter side. #173


cloudwebrtc:
> I signed it!

stuartmorgan:
Sorry for the long initial delay in reviewing; for some reason I had been thinking this was gated on the engine texture change, rather than a stand-alone piece that could land at any time.

PR #182, opened 2018-12-26
[macOS] Align FLEPlugin with iOS Flutter

stuartmorgan:
Changes FLEPlugin to be a subset of iOS's FlutterPlugin's API, including
changing the plugin creation and registration system. Plugins now create
and manage their own channels, instead of relying on FLEViewController
for high-level message APIs.

This is a breaking change for any FLEPlugin implementation, as well as
application-level code that was calling addPlugin:.

This essentially completes issue #102 for macOS.


library/macos/FLEViewController.m
```diff
@@ -219,53 +234,11 @@ - (BOOL)launchEngineWithMainPath:(NSURL *)main
...
-
-- (void)invokeMethod:(NSString *)method arguments:(id)arguments onChannel:(NSString *)channel {
-  [self invokeMethod:method
-           arguments:arguments
-           onChannel:channel
-           withCodec:[FLEJSONMethodCodec sharedInstance]];
+- (id<FLEPluginRegistrar>)registrarForPlugin:(NSString *)pluginName {
+  // Currently, the view controller acts as the registrar for all plugins, so the
+  // name is ignored. It is part of the API to reduce churn in the future when
+  // aligning more closely with the Flutter registrar system.
```
krisgiesing: Out of curiosity, what does the Flutter registrar system do with the name?

library/macos/FLEPlugin.h
```diff
@@ -18,52 +18,38 @@
...
-@optional
-
-/**
- * If implemented, returns the codec to use for method calls to this plugin.
- *
- * If not implemented, the codec is assumed to be FLEJSONMethodCodec. Note that this is different
- * from existing Flutter platforms, which default to the standard codec; this is to preserve
- * backwards compatibility for FLEPlugin until the breaking change for the platform messages API.
- */
-@property(nonnull, readonly) id<FLEMethodCodec> codec;
```
krisgiesing: I guess this is no longer needed because the type of codec is now private to the plugin and doesn't need to be viewed by anything outside?

stuartmorgan: More than that, there's no longer necessarily such thing as "the type of codec" for a plugin. A plugin can have any number of channels it wants (including zero, theoretically), and they don't all necessarily have to use the same codec.

library/macos/FLEViewController.m
stuartmorgan: Registrars are one-per-plugin in Flutter, so the name is used as a map key for finding the correct registrar.

library/macos/FLEViewController.m
```diff
@@ -157,6 +159,19 @@ static bool OnMakeResourceCurrent(FLEViewController *controller) {
...
 @implementation FLEViewController {
   FlutterEngine _engine;
+
+  // The additional context provided to the Flutter engine for resource loading.
+  NSOpenGLContext *_resourceContext;
+
+  // A mapping of channel names to the registered handlers for those channels.
+  NSMutableDictionary<NSString *, FLEBinaryMessageHandler> *_messageHandlers;
+
+  // The plugin used to handle text input. This is not an FLEPlugin, so much be owned separately.
```
franciscojma86: s/much/must

stuartmorgan: Done.

example/macos/ExampleWindow.swift
```diff
@@ -18,9 +18,12 @@ class ExampleWindow: NSWindow {
   @IBOutlet weak var flutterViewController: FLEViewController!
 
   override func awakeFromNib() {
-    flutterViewController.add(FLEColorPanelPlugin())
-    flutterViewController.add(FLEFileChooserPlugin())
-    flutterViewController.add(FLEMenubarPlugin())
+    FLEColorPanelPlugin.register(
+      with: flutterViewController.registrar(forPlugin: "FLEColorPanelPlugin"))
+    FLEFileChooserPlugin.register(
```
sassman: the wiki seem to contain only the usage in ObjectiveC, this I guess this code shows correctly how to use the file chooser plugin for instance

stuartmorgan: > the wiki seem to contain only the usage in ObjectiveC

Calling plugin APIs from Swift is exactly the same as calling any other ObjC code from Swift.

PR #183, opened 2018-12-26
[linux] Align C++ plugin structure with Flutter

stuartmorgan:
Eliminates Plugin as the primary interface point, and instead moves to a
MethodChannel-based model. This eliminates the need for void* types in
the interfaces, allowing a move to full type safety (via templates) for
all the classes related to channels.

Additionally, the plugin creation and registration system now matches
the Flutter structure on iOS and Android, using static registration of a
plugin class with a PluginRegistrar. The desktop registrar currently has
minimal functionality, but it can be extended over time.

This is a breaking change for any Plugin implementation, as well as
application-level code that was calling AddPlugin. It is collapsed into
one large patch in order to minimize the number of API-breaking changes.

This essentially completes issue #102 for Linux.


stuartmorgan:
Note: please wait for me to land this rather than landing it once it's approved, because it's a breaking change for PR #161 and I have a patch I'll want to push there at the same time as landing this.

stuartmorgan:
Rebased to account for 161 (now depends on #194)

plugins/color_panel/linux/src/color_panel_plugin.cc
```diff
@@ -89,13 +90,37 @@ class ColorPanelPlugin::ColorPanel {
...
+void ColorPanelPlugin::RegisterWithRegistrar(
+    flutter_desktop_embedding::PluginRegistrar *registrar) {
+  auto channel =
+      std::make_unique<flutter_desktop_embedding::MethodChannel<Json::Value>>(
+          registrar->messenger(), kChannelName,
+          &flutter_desktop_embedding::JsonMethodCodec::GetInstance());
+  auto *channel_pointer = channel.get();
+
+  std::unique_ptr<ColorPanelPlugin> plugin(
+      new ColorPanelPlugin(std::move(channel)));
```
franciscojma86: Would it be better to use make_unique here?

library/include/flutter_desktop_embedding/engine_method_result.h
```diff
@@ -0,0 +1,90 @@
...
+
+#include "binary_messenger.h"
+#include "method_codec.h"
+#include "method_result.h"
+
+namespace flutter_desktop_embedding {
+
+namespace internal {
+// Manages the one-time sending of response data. This is an internal helper
+// class for EngineMethodResult, separated out since the implmentation doesn't
```
franciscojma86: implementation*

library/common/internal/plugin_handler.cc
```diff
@@ -94,4 +84,14 @@ void PluginHandler::SetMessageHandler(const std::string &channel,
   handlers_[channel] = std::move(handler);
 }
 
+// PluginRegistrar:
+
```
franciscojma86: Remove extra line?

library/include/flutter_desktop_embedding/engine_method_result.h
stuartmorgan: Done.

library/common/internal/plugin_handler.cc
stuartmorgan: It's a heading for both of the following methods, so I separated it so that it doesn't appear to be associated only with AddPlugin.

plugins/color_panel/linux/src/color_panel_plugin.cc
stuartmorgan: Unfortunately make_unique doesn't work with classes whose constructors are private; see https://abseil.io/tips/134

The question is a good reminder to a comment as the tip suggests, so I've done that :)

library/common/glfw/text_input_plugin.h
```diff
@@ -19,27 +19,22 @@
...
-  TextInputPlugin();
+  TextInputPlugin(PluginRegistrar *registrar);
   virtual ~TextInputPlugin();
 
-  // Plugin.
-  void HandleJsonMethodCall(const JsonMethodCall &method_call,
-                            std::unique_ptr<MethodResult> result) override;
-
-  // KeyboardHookHandler.
+  // KeyboardHookHandler:
```
franciscojma86: Should've added this to my previous comment. I asked to remove the extra line since there's no line here. Maybe add one here? :)

stuartmorgan: This is a style I picked up from Chromium (IIRC): all the overridden function declarations from one superclass are declared as a blob, without blank lines or comments (since the declaration comment is at the original declaration). Since it's one blob, the comment is attached to it without a blank line.

In the implementation file, on the other hand, there are always blank lines between function definitions, so the comment needs to be separated so it doesn't appear to be an implementation comment for one function. Maybe doing the implementation header is confusing in C++, but I find it helpful in navigating ObjC files (where it's very common, using #pragma mark) so I thought I'd give it a try here.

franciscojma86: Makes sense, thanks!

PR #184, opened 2018-12-27
[linux] Add GLFW-based proc resolving

### LINKED ISSUE #170: [Linux] Failed to setup Skia Gr context in some configurations
With recent builds, some people are seeing persistent "Failed to setup Skia Gr context" logging + crash on startup.

Unlike the previous instance (#135), this does not affect all Linux configurations, only some, which suggests this may be driver related, or capability-related. For instance, in at least one instance I'm aware of the crash happens when using a machine normally, but not when using Chrome Remote Desktop to access the same machine.

Things that could be useful from people experiencing this:

1. Information about your machine configuration
2. Testing with older versions of Flutter to try to find a regression range (since this seems to have started recently for people, I expect this was triggered by a change in the Flutter engine).
3. Building a local Flutter engine and debugging the failure

(3) would be the most useful, but is obviously the most involved. I believe @franciscojma86 will be trying this route.

GeertJohan:
Fixes #170 

The default symbol resolver that is used when no specific resolver is given doesn't resolve according to OpenGL features available in the GLFW context. This causes errors for some users.


stuartmorgan:
It seems like this is patching over an issue with the engine fix for #135. @chinmaygarde any thoughts on why some configurations are still not working without this? Can it be fixed in the engine?

library/linux/src/embedder.cc
```diff
@@ -199,6 +199,12 @@ static void GLFWClearCanvas(GLFWwindow *window) {
   glfwMakeContextCurrent(nullptr);
 }
 
+// Resolves the address of the specified OpenGL or OpenGL ES 
+// core or extension function, if it is supported by the current context.
+static void* GLFWProcResolver(void *user_data, const char *name) {
+  return (void*) glfwGetProcAddress(name);
```
stuartmorgan: If we do end up needing this fix on the embedding side, this needs to be `static_cast<void*>(...)`. This project follow's Google's C++ style guide, which does not use C-style casts.

GeertJohan: I only could get it to work with `reinterpret_cast`. But please double check what I'm doing because I'm not very familiar with C++.

stuartmorgan: I'll need to look at this again when I have more time. Having looked at the types, this is reinterpret-casting a function pointer to a void*, which I believe violates the spec—in which either the engine API is problematic, or there's a missing step here.

stuartmorgan: I dug into this more to see what Flutter code this was ultimately interacting with and how they handle the mismatch. It turns out that it eventually comes down to `dlsym`, so this is the issue described in the Rationale section [here](http://pubs.opengroup.org/onlinepubs/009695399/functions/dlsym.html). Pushing the requirement for the non-standards-compliant conversion up to this abstract layer, rather than keeping it closer to the `dlsym` that requires POSIX, does seem like an engine API problem.

However, since the bug is breaking a bunch of people, I'm going to go ahead and land this, and then follow up with the Flutter team separately about whether there should be more changes here at the engine level. Thanks for tracking down the issue!

PR #187, opened 2018-12-28
[glfw] Add pixel_ratio to the embedder

GeertJohan:
Related to #32 

This PR adds pixel_ratio calculations to the linux/GLFW embedder.

The pixel_ratio seems to be based on 160 dpi. This is the same conversion rate for the `dp` in Android. The Flutter documentation mentions the Nexus 6 to have a pixel_ratio of 3.5, which matches with it's pixels/dp scale-factor as well.

GLFW makes use of screen coordinates, which are a virtual unit and do not always map 1:1 to physical pixels on a device. I am not 100% sure if the conversion to actual pixels is needed, or if Flutter should just know about screen coordinates?


stuartmorgan:
I don't think we need the DPI calculations here; on macOS we just pass the pixel scale factor (e.g., 2 for Retina displays), so `pixelsPerScreenCoordinate` is likely the relevant value in your code. However, I'm not sure what, if anything, you need to do in GLFW to actually cause the GLFW window to use a high-resolution buffer though, or if that will be done automatically. Are you seeing a value other than 1 there without any other changes?

One you provide a non-1 value for `pixel_ratio` you'll also need to adjust any coordinates sent to Flutter, or click handling will break.

GeertJohan:
This was my initial thought as well. But if I used just the pixelsPerScreenCoordinate, the application ofcourse still renders tiny (text barely readable). On my display, 1 pixel = 1 screen coordinate, so nothing changes. I believe that 1:1 ratio is the "normal" case for most linux users; there is no up-scaling performed by the GPU/driver and applications get the full display (in my case 4K) to render. In those cases the ratio would always be `1.0`.

After digging into the inner workings of Flutter I found that the logical pixels, as given by the MediaQuery are actually the physical pixels we pass here, divided by the ratio we pass here. Logical pixels are used for rendering things equally across devices.

> Logical pixels are roughly the same visual size across devices. 
> https://docs.flutter.io/flutter/widgets/MediaQueryData/size.html

For example:
A Nexus 6 has a pixel_ratio of 3.5 (560 DPI of the device divided by 160 as baseline 1.0 scale).
It's physical resolution is 1440x2560. But the Flutter MediaQuery size gives us only 411.4x683.4 logical pixels.

If I create a Container widget at 400 width, it is rendered at 400*3.5=1400 physical pixels on the Nexus 6. And it is rendered at 400*1.6=640 pixels on my laptop. The physical size in inches is nearly equal on both devices, which I believe is the goal of Flutter.

Therefore I think we must take the DPI of the display into account here, so that the logical pixels in Flutter are calculated correctly on any device.

The one thing I'm not sure of is screen coordinates vs pixels in GLFW. If Flutter renders to a canvas of screen coordinates, we should not calculate and apply pixelsPerScreenCoordinates. I would break things (render out of frame, or only use a small part of the frame). If Flutter renders to actual physcial pixels (as defined on the framebuffer), then we should.

> One you provide a non-1 value for pixel_ratio you'll also need to adjust any coordinates sent to Flutter, or click handling will break.

On my machine I'm running with 1.6, and click handling works as before. But pixelsPerScreenCoordinates is 1.0, so there is no difference between the width/height we pass to flutter here and the coordinates obtained from `glfwGetCursorPos`.
Indeed, if pixelsPerScreenCoordinates is actually needed, then we should apply that factor to positions in `GLFWmouseButtonCallback`.

I will be able to test on multiple monitors soon. I hope that gives me a non-1.0 pixelsPerScreenCoordinates to see how that works out for the rendering.

stuartmorgan:
> I believe that 1:1 ratio is the "normal" case for most linux users; there is no up-scaling performed by the GPU/driver and applications get the full display (in my case 4K) to render.

I see, it sounds like I'm overfitting from the way macOS handles high-resolution displays then.

> If Flutter renders to actual physcial pixels (as defined on the framebuffer)

Unless the engine works very differently on Linux than macOS (which seems unlikely for the layer that would be involved), the sizes and coordinates passed through the embedding API should be in framebuffer coordinates. E.g., `glfwGetFramebufferSize`'s output values should be what are passed as the width and height in that struct AFAIK. The current GLFW embedder code relies on the multiplier being 1.

stuartmorgan:
Pushed a merge, since my refactoring to combine the Linux and Windows implementation conflicted with this. As a result, this now improves the Windows implementation as well :)

library/common/glfw/embedder.cc
```diff
@@ -67,41 +78,91 @@ static FlutterEmbedderState *GetSavedEmbedderState(GLFWwindow *window) {
...
+  state->monitorHeightMm = monitorHeightMm;
+
+  auto monitorMode = glfwGetVideoMode(monitor);
+  state->monitorWidth = monitorMode->width;
+  state->monitorHeight = monitorMode->height;
+
+  SyncWindowMetrics(window);
+}
+
+static void GLFWWindowSizeCallback(GLFWwindow *window, int width, int height) {
```
stuartmorgan: This should be eliminated; GLFWFramebufferSizeCallback is called for resizes as well, so currently the code is updating the engine twice for every resize. You can just call `glfwGetWindowSize` in the framebuffer callback to compute the pixels-per-coordinate ratio.

library/common/glfw/embedder.cc
```diff
@@ -304,10 +365,20 @@ GLFWwindow *CreateFlutterWindow(size_t initial_width, size_t initial_height,
 
   AddPlugin(window, std::move(input_plugin));
 
+  UpdateMonitorStats(window);
+
   int width, height;
   glfwGetWindowSize(window, &width, &height);
-  GLFWwindowSizeCallback(window, width, height);
-  glfwSetWindowSizeCallback(window, GLFWwindowSizeCallback);
+  glfwSetWindowSizeCallback(window, GLFWWindowSizeCallback);
```
stuartmorgan: FWIW, the lower-case W here was intentional, since the typename of the thing being resized is `GLFWwindow` rather than `GLFWWindow`; since it's being removed it doesn't really matter though. The mouse/cursor capitalization looks like a mistake though.

library/common/glfw/embedder.cc
```diff
@@ -57,6 +57,17 @@ struct FlutterEmbedderState {
   // Handles raw key interactions from GLFW.
   // TODO: Revisit ownership model once Issue #102 is resolved.
   std::unique_ptr<flutter_desktop_embedding::KeyEventHandler> key_event_handler;
+
+  // window dimensions, in screen coordinates.
+  int width, height = 1;
+  // window framebuffer dimensions, in pixels.
+  int widthPx, heightPx = 1;
```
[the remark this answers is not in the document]
GeertJohan: I will remove the calculated value (`pixelsPerScreenCoordinate`) to avoid making extra function calls. Local math is a lot cheaper.

library/common/glfw/embedder.cc
```diff
@@ -67,41 +78,91 @@ static FlutterEmbedderState *GetSavedEmbedderState(GLFWwindow *window) {
...
+
+  FlutterWindowMetricsEvent event = {};
+  event.struct_size = sizeof(event);
+  event.width = state->widthPx;
+  event.height = state->heightPx;
+  event.pixel_ratio = pixel_ratio;
+  FlutterEngineSendWindowMetricsEvent(state->engine, &event);
+}
+
+static void UpdateMonitorStats(GLFWwindow *window) {
```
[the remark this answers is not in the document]
GeertJohan: This function only has one caller and very low complexity. Will inline it into `CreateFlutterWindow`.

library/common/glfw/embedder.cc
GeertJohan: I changed the capitalization because it was inconsistent across GLFW* functions. `GLFWKeyCallback` is already with capital K. Which should it be?

GeertJohan: I would argue that it should have been capital W though, because that's just the name of the callback in GLFW.
If we wanted to include the typename of the thing that's on which the callback is applied consistently it should imo be `GLFWwindowWindowSizeCallback`, `GLFWwindowFramebufferSizeCallback`, `GLFWwindowMouseButtonCallback` etc.

GeertJohan: We could actually still do that right now? Prefixing all the callback functions with the type on which they are called (`GLFWwindow`) instead of just `GLFW`.

library/common/glfw/embedder.cc
GeertJohan: Removed

library/common/glfw/embedder.cc
```diff
@@ -67,41 +78,91 @@ static FlutterEmbedderState *GetSavedEmbedderState(GLFWwindow *window) {
...
+  int monitorWidthMm, monitorHeightMm;
+  glfwGetMonitorPhysicalSize(monitor, &monitorWidthMm, &monitorHeightMm);
+  state->monitorWidthMm = monitorWidthMm;
+  state->monitorHeightMm = monitorHeightMm;
+
+  auto monitorMode = glfwGetVideoMode(monitor);
+  state->monitorWidth = monitorMode->width;
+  state->monitorHeight = monitorMode->height;
+
+  SyncWindowMetrics(window);
```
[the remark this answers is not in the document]
GeertJohan: Indeed, this is fixed now.

GeertJohan:
Thanks for the feedback @stuartmorgan. I have pushed the changes. There's one conversation still open/unclear to me, about the `GLFWwindow` prefix. When that is resolved I hope this should be good to merge.

library/common/glfw/embedder.cc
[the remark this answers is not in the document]
stuartmorgan: There's enough going on in CreateFlutterWindow that I would prefer monitor detail calculation to stay factored out into a function.

library/common/glfw/embedder.cc
```diff
@@ -57,6 +57,13 @@ struct FlutterEmbedderState {
   // Handles raw key interactions from GLFW.
   // TODO: Revisit ownership model once Issue #102 is resolved.
   std::unique_ptr<flutter_desktop_embedding::KeyEventHandler> key_event_handler;
+
+  // Window framebuffer dimensions, in pixels.
+  int width_px, height_px = 1;
```
stuartmorgan: Why are these statefull instead of passed as parameters? It looks like they are only ever stored just before calling a function and read inside that function.

GeertJohan: Check, I have inlined SyncWindowMetrics.

GeertJohan:
The commit history kinda became a mess with all the changes, so I have rebased the changes on the latest master and opened a new PR for it. #204

stuartmorgan:
For future reference we land all PRs with squash-and-merge, so the commit history of PRs doesn't matter.

PR #194, opened 2019-01-02
Merge GLFW library implementations

stuartmorgan:
Merges the GLFW-based embedding implementations for Windows and
Linux into a single shared implementation. At this point the implementations
are largely identical, and having two copies is increasingly a maintenance
burden.

Originally they were separate since they were intended to be throw-away,
and their development proceeded at different speeds. Now that they have
reached parity, and there's a possibility that the GLFW implementation
will be longer-lived than initially expected, having one copy that's testable
on several platforms should simplify development.


library/common/glfw/embedder.cc
```diff
@@ -26,6 +27,20 @@
 #include "library/common/glfw/text_input_plugin.h"
 #include "library/common/internal/plugin_handler.h"
 
+#ifdef __linux__
```
krisgiesing: Can you add a comment explaining why these includes are here? (If the dependencies change in the future it's nice to have a reminder to clean this up.)

stuartmorgan: Done.

PR #198, opened 2019-01-03
Initial Travis configuration

stuartmorgan:
Builds Linux via Make and GN, and builds macOS.

### DIFF
--- a/.travis.yaml
+++ b/.travis.yaml
@@ -1,0 +1,35 @@
+matrix:
+  include:
+    - os: linux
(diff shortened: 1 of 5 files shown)


stuartmorgan:
This may not end up being a long-term solution, but if it's easy to get partial compilation coverage of PRs that would help.

(There are almost certainly bugs here, since I've never used Travis before, but it seems that it won't run on PRs until there's a config checked in, so the first land has to be untested AFAICT.)

PR #199, opened 2019-01-03
Update release mode flag

stuartmorgan:
The Flutter engine renamed the flag that controls the "Debug" banner
as part of the Dart 2 transition (Engine PR 6977). Update the example
apps to pass the new flag in release mode.


stuartmorgan:
We may want to change this to be internal to the framework, rather than a string that each embedder implementation has to manage, but IIRC there are plans to change the Engine's argument API so I'll wait to see what happens there before making structural changes.

PR #201, opened 2019-01-04
Fix Travis config

stuartmorgan:
- Fix the filename, so Travis is actually active.
- Various fixes to make the builds green.


stuartmorgan:
It turns out that the reason Travis wasn't running on the initial PR wasn't that it had to be landed on master first, but that the filename was wrong. Doh!

But as expected, there were other issues as well.

Issue #160, opened 2018-11-30
Add a clean way to use a custom Flutter engine


stuartmorgan:
Currently the only way to do this is to do a normal build, then replace the library/framework with the locally built version (and remember that your build is no longer actually using the hash that `.last_engine_version`, and thus the build output, claims). Updating Flutter will stomp the replacement.

And you have to also remember to change
https://github.com/google/flutter-desktop-embedding/blob/master/tools/build_flutter_assets#L26
and undo it when unwinding.

There should be something along the lines of `.flutter_location_config` that both `update_flutter_engine` callers and `build_flutter_assets` will check against that will cause a local build to be copied in and used instead, such that correctly enabling/disabling it is as simple as moving a file. (Or, if we switch to GN, a GN configuration option might work instead.)

PR #210, opened 2019-01-07
Add a generic run_dart_tool command

stuartmorgan:
This is a generic version of the existing update_flutter_engine wrappers,
so that as more dart tools are added they don't each need a wrapper.

update_flutter_engine remains for now, but may be removed in the future
in favor of calling 'run_dart_tool update_flutter_engine' instead. The shell
script version has been changed to match the existing batch script
behavior of automatically providing --flutter_root (since now there's an
easy way for a caller who wants to provide something else to do so).
This may simplify future changes to provide a project-level engine
override (#160)


stuartmorgan:
It's used in all our existing batch files; I was assuming it was relatively standard. But since I'm new to batch scripting, I don't really have any context. (It seems much less crazy to me that using `for` to set a variable!)

Once things settle a bit with the new policy to minimize batch/shell in favor of Dart, I'll see how often they need to be touched. Hopefully they will need little to no ongoing maintenance since they will be so small, but if not I can add a bunch of comments to make it easier to maintain them regardless of familiarity level.

PR #211, opened 2019-01-07
Dart analysis-related fixes

stuartmorgan:
- Fixes the include path for the example app's analysis file, which
  broke during the repository restructure.
- Removes analysis options that are obsolete in recent Dart versions.
- Fix some new analysis issues (mostly constructor ordering).

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -4,10 +4,6 @@
 
 # Root analysis options shared among all Dart code in the respository. Based
 # on the Fuchsia standard analysis options, with some changes.
-analyzer:
-  strong-mode: true
-  language:
-    enableSuperMixins: true
 linter:
   # Full list available at http://dart-lang.github.io/linter/lints/options/options.html.
   rules:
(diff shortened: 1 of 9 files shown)


analysis_options.yaml
```diff
@@ -4,10 +4,6 @@
 
 # Root analysis options shared among all Dart code in the respository. Based
 # on the Fuchsia standard analysis options, with some changes.
-analyzer:
```
krisgiesing: FWIW, there are still strong mode options that aren't the default (see implicit-cast and implicit-dynamic). We're considering trying to turn those on for Variant, but they trigger a lot of analysis failures right now.

krisgiesing: https://www.dartlang.org/guides/language/analysis-options

stuartmorgan: Thanks for the pointer! I tried turning them on, and fixed some of the more obvious errors (added to this PR), but some of them were more fiddly, and there aren't easy-to-find example solutions for some of them, so I'm leaving them off for now.

PR #219, opened 2019-01-08
[linux/windows] Add BasicMessageChannel

stuartmorgan:
Implements BasicMessageChannel, adds MessageCodec, and makes
JsonMessageCodec public.

Converts the key event handler from using BinaryMessenger directly to
using a BasicMessageChannel to send events now that it's available.

### DIFF
--- a/library/include/flutter_desktop_embedding/basic_message_channel.h
+++ b/library/include/flutter_desktop_embedding/basic_message_channel.h
@@ -1,0 +1,106 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


library/include/flutter_desktop_embedding/basic_message_channel.h
```diff
@@ -0,0 +1,106 @@
...
+        codec_->EncodeMessage(message);
+    messenger_->Send(name_, raw_message->data(), raw_message->size());
+  }
+
+  // TODO: Add support for a version of Send expecting a reply once
+  // https://github.com/flutter/flutter/issues/18852 is fixed.
+
+  // Registers a handler that should be called any time a message is
+  // received on this channel.
+  void SetMessageHandler(MessageHandler<T> handler) const {
```
awdavies: I'm a bit confused about what this does.  It doesn't look like it gets called anywhere? Is this just due to the implementation of the key event handler/

stuartmorgan: Right, the key event handler is send-only, so never does this. I just didn't want to add MessageChannel without half of its functionality.

It does the same thing SetMethodCallHandler does on MethodChannel, just without the extra structure assumed on the message.

PR #221, opened 2019-01-09
Remove exports.def from GN Build

CallumIddon:
GN build fails as `exports.def` cannot be found.

### DIFF
--- a/library/BUILD.gn
+++ b/library/BUILD.gn
@@ -90,11 +90,6 @@ published_shared_library("flutter_embedder") {
   }
 
   if (is_win) {
-    dll_exports = rebase_path("windows/exports.def")
-    ldflags = [
-      "/DEF:$dll_exports",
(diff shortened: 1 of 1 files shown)


stuartmorgan:
> Maybe I pushed the branch with the fix as an uncommitted local change.

This is exactly what happened :( Sorry for the entirely unnecessary bustage.

(I'd like to get CI set up for Windows to help avoid this kind of thing, but haven't had a chance yet.)

CallumIddon:
> I'd like to get CI set up for Windows to help avoid this kind of thing, but haven't had a chance yet.

I am happy to tackle this if you havn't already gotten started on it yet. How are you thinking we install the Microsoft Visual Studio Build Tools?

stuartmorgan:
That would be great! I'll file a bug for tracking/discussion and tag you.

PR #225, opened 2019-01-10
Add a simpler way to use a local Flutter engine

### LINKED ISSUE #160: Add a clean way to use a custom Flutter engine

stuartmorgan:
update_flutter_engine and build_flutter_assets now check for an
engine_override file at the root of the repository to trigger the
use of a local engine. Updates the README to link to instructions
on using it.

Since this adds more logic to build_flutter_assets, it has been
rewritten in Dart.

There are some obvious limitations (e.g., hard-coded name/location
requirement, copying every time instead of only when the
local build has changed), which can be addressed in the future if
there's sufficient demand. Since it is only intended for temporary use
when debugging, or testing engine changes, this may be sufficient,
and is a significant improvement over managing it manually.

Unlike .flutter_location_config this has no leading ., and is in the
repository rather than above it. This difference is intentional,
since unlike .flutter_location_config this is for temporary use,
so having it show up in directory listings and git status is
helpful as a reminder that it is enabled.

Fixes #160

### DIFF
--- a/LocalEngine.md
+++ b/LocalEngine.md
@@ -1,0 +1,35 @@
+# Running a Locally Built Flutter Engine
+
+It may be useful for debugging or development to run with a [locally built
+Flutter engine](https://github.com/flutter/flutter/wiki/Compiling-the-engine).
+To temporarily override the normal behavior of `update_flutter_engine` and
+`build_flutter_assets`, add a file called `engine_override` at the root of
+your `flutter_desktop_embedding` checkout. For instance, on macOS or Linux:
+```
+$ echo host_debug_unopt > engine_override
+```
+
+This will cause `update_flutter_engine` to copy your local engine instead of
+downloading a prebuilt engine, and `build_flutter_assets` to pass the
+`--local-engine` flag when building assets.
+
+**Important**: Your Flutter engine checkout must be in a folder called `engine`
+(as recommended in the [setup
+instructions](https://github.com/flutter/flutter/wiki/Setting-up-the-Engine-development-environment)),
+which must be next to your `flutter` checkout, or `update_flutter_engine` will
+fail.
+
+## Dart Changes
+
+If you change any Dart code in your engine, be sure to follow the instructions
+[here](https://github.com/flutter/flutter/wiki/Setting-up-the-Engine-development-environment)
+or
+[here](https://github.com/flutter/flutter/wiki/The-flutter-tool#using-a-locally-built-engine-with-the-flutter-tool)
+for adding a `dependency_overrides` section to your `pubspec.yaml`.
+
+## Switching Back to a Prebuilt Engine
+
+To stop using your local engine, just delete `engine_override` (and
+`dependency_overrides` if you added them to `pubspec.yaml`) and rebuild.
+`update_flutter_engine` will automatically re-download the correct prebuilt
+engine.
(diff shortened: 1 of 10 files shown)


stuartmorgan:
Let me know if you'd prefer having the first commit here (converting `build_flutter_assets` to Dart, without any functional changes) pulled out into a separate PR for reviewing.

krisgiesing:
When I did this I also had to modify the Xcode projects to look in my override location for frameworks (i.e. I had to override the framework paths setting in the project). Does this approach avoid the need for that?

stuartmorgan:
Yes, with this approach `update_flutter_engine` switches to being a local copy operation instead of a download operation when the override file is present.

So for instance, in the example app Xcode runs `update_flutter_engine` on every build, pointing at `flutter_engine_framework/macos`. Normally this does a hash check against the engine version listed in the flutter tree, and is either a no-op or a download if it's changed. If you add `engine_override`, that same step will instead copy the local build to `flutter_engine_framework/macos` (bypassing the hash check, and updating `.last_engine_version`. Since it's in the same location, nothing else about the build process needs to change.

(In practice, I was doing a manual version of this—leaving the hash unchanged so it wouldn't re-download and stomp my copied version—rather than adjusting paths in the project, but it was somewhat error-prone.)

This is more inefficient than modifying the project (extra copy operations, an extra download of the prebuilt version when it's turned back off), but it's very simple to use, and it works exactly the same way regardless of platform or build system (which in turn makes it easy to document).

LocalEngine.md
```diff
@@ -0,0 +1,35 @@
+# Running a Locally Built Flutter Engine
+
+It may be useful for debugging or development to run with a [locally built
+Flutter engine](https://github.com/flutter/flutter/wiki/Compiling-the-engine).
+To temporarily override the normal behavior of `update_flutter_engine` and
+`build_flutter_assets`, add a file called `engine_override` at the root of
```
franciscojma86: Just curious, why isn't this like .flutter_location_config? Passing the path to the engine as opposed to be forced to checkout the engine in a specific place?

stuartmorgan: Just a benefit/effort tradeoff. `.flutter_location_config` allows for greater flexibility in the regular build environment, where there can be other constraints on the relative location of dependencies, whereas using `engine_override` is only meant for temporary local use (if a project wants to use a custom engine as a standard behavior, they just shouldn't use `update_flutter_engine` in their build for instance). So I think that where the engine checkout is located is much less likely to be an issue—especially since the Engine build docs themselves recommend using a folder called `engine` next to the Flutter tree.

Given that, writing the path configuration in Dart (making sure it handled relative paths, absolute paths, `~/...`, etc.) and testing all the variations didn't seem worth the investment of time at the moment.

If someone ends up needing that configurability, patches welcome at that point :)

franciscojma86: Makes sense, thanks!

PR #235, opened 2019-01-16
Updated bundle ID for FlutterEmbedder.framework

jbg:
Fix: https://github.com/google/flutter-desktop-embedding/issues/234

Update the bundle ID for FlutterEmbedder.framework to match the change at https://github.com/flutter/engine/pull/7503


jbg:
I signed it!

jbg:
added support for loading with the old ID if loading with the new ID fails.

PR #236, opened 2019-01-16
Simplify the API for launching Flutter

stuartmorgan:
- Eliminate the distinction between source mode and snapshot mode;
  source mode hasn't worked since the switch to Dart2, and the relevant
  engine arguments are deprecated in
  https://github.com/flutter/engine/pull/7497
- Hide the need for the first argument to be the executable from clients
  of the library; this is an implementation detail of the engine API,
  and not something that each client application should need to handle
  given that the expected use case is to provide a constructed argument
  array, rather than pass through argc/argv directly.
  (As part of this change, removes the macOS code to strip Xcode-added
  arguments, for the same reason.)

This is a breaking change; clients will need to update the call to the
primary entry point, including removing the dummy first argument they
were likely providing in the arguments array.

### DIFF
--- a/library/common/glfw/embedder.cc
+++ b/library/common/glfw/embedder.cc
@@ -250,7 +250,7 @@ static FlutterEngine RunFlutterEngine(
     const std::string &icu_data_path,
     const std::vector<std::string> &arguments) {
   // FlutterProjectArgs is expecting a full argv, so when processing it for
-  // flags the first item treated as the executable and ignored. Add a dummy
+  // flags the first item is treated as the executable and ignored. Add a dummy
   // value so that all provided arguments are used.
   std::vector<const char *> argv = {"placeholder"};
   std::transform(
--- a/library/common/glfw/embedder.cc
+++ b/library/common/glfw/embedder.cc
@@ -246,11 +246,13 @@ static void *GLFWProcResolver(void *user_data, const char *name) {
 //
 // Returns a caller-owned pointer to the engine.
 static FlutterEngine RunFlutterEngine(
-    GLFWwindow *window, const std::string &main_path,
-    const std::string &assets_path, const std::string &packages_path,
+    GLFWwindow *window, const std::string &assets_path,
     const std::string &icu_data_path,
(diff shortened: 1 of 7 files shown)


library/common/glfw/embedder.cc
```diff
@@ -246,11 +246,13 @@ static void *GLFWProcResolver(void *user_data, const char *name) {
...
 // Returns a caller-owned pointer to the engine.
 static FlutterEngine RunFlutterEngine(
-    GLFWwindow *window, const std::string &main_path,
-    const std::string &assets_path, const std::string &packages_path,
+    GLFWwindow *window, const std::string &assets_path,
     const std::string &icu_data_path,
     const std::vector<std::string> &arguments) {
-  std::vector<const char *> argv;
+  // FlutterProjectArgs is expecting a full argv, so when processing it for
+  // flags the first item treated as the executable and ignored. Add a dummy
```
franciscojma86: s/and/is ?

library/macos/FLEViewController.m
```diff
@@ -280,38 +263,30 @@ - (BOOL)launchEngineInternalWithAssetsPath:(NSURL *)assets
...
-  for (int i = 0; i < argc; ++i) {
-    if ([arguments[i] isEqualToString:kXcodeExtraArgumentOne] && (i + 1 < argc) &&
-        [arguments[i + 1] isEqualToString:kXcodeExtraArgumentTwo]) {
-      unused += 2;
-      ++i;
-      continue;
-    }
-    command_line_args[i - unused] = [arguments[i] UTF8String];
+  // FlutterProjectArgs is expecting a full argv, so when processing it for flags the first
+  // item treated as the executable and ignored. Add a dummy value so that all provided arguments
```
franciscojma86: same here

franciscojma86:
LGTM with with two typo comments

library/common/glfw/embedder.cc
stuartmorgan: Good catch; should have been "the first item **is** treated as the executable and ignored". Fixed in both files.

Issue #47, opened 2018-04-24
Stop using internal system channel for text input handling


stuartmorgan:
The current text input plugin is based on implementation details of the engine that are not public/stable API. Once there's a clearly defined text input API for embedders we need to switch to using that instead of the current unsupported approach.

krisgiesing:
Can you elaborate on which specific implementation details of the engine are relevant?

stuartmorgan:
My understanding is that essentially everything in the text input system as currently implemented: the platform channel names of the messages the plugin handles, the keys and structures in the response objects, and in general the whole concept of implementing text input through the generic platform messaging system rather than a formal API.

krisgiesing:
Ah, I think you mean "framework" then rather than "engine"? That is, these are Dart constructs we're referencing, not C/C++ constructs.

stuartmorgan:
Closing; this is a non-issue now that the shells are part of flutter/engine. (There should probably still be a solution to this for other embedders, but that's no longer in scope for the desktop work.)

PR #239, opened 2019-01-19
add support for shift, option (alt), control and command (meta) modifier keys

jbg:
Passes through the `NSEventModifierFlag` values for shift, option, control and command as Android `metaState` flags.

### DIFF
--- a/library/macos/FLEViewController.m
+++ b/library/macos/FLEViewController.m
@@ -34,10 +34,10 @@ static NSString *const kICUBundlePath = @"icudtl.dat";
 static const int kDefaultWindowFramebuffer = 0;
 
 // Android KeyEvent constants from https://developer.android.com/reference/android/view/KeyEvent
-static const int kAndroidMetaStateShift = 1;
-static const int kAndroidMetaStateAlt = 2;
-static const int kAndroidMetaStateCtrl = 4096;
-static const int kAndroidMetaStateMeta = 65536;
+static const int kAndroidMetaStateShift = 1 << 0;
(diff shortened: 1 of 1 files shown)


library/macos/FLEViewController.m
```diff
@@ -33,6 +33,12 @@
 
 static const int kDefaultWindowFramebuffer = 0;
 
+// Android KeyEvent constants from https://developer.android.com/reference/android/view/KeyEvent
+static const int kAndroidMetaStateShift = 1;
+static const int kAndroidMetaStateAlt = 2;
+static const int kAndroidMetaStateCtrl = 4096;
+static const int kAndroidMetaStateMeta = 65536;
```
stuartmorgan: I'd prefer to use the more common convention of expressing bit flags via shifting, rather than as decimal numbers, since it more clearly expresses how they are used. E.g., `...StateMeta = 1 << 16`.

jbg: Done.

krisgiesing:
Why are we using Android-defined constants here? Shouldn't we use a platform-independent description of the modifier key state?

stuartmorgan:
> Shouldn't we use a platform-independent description of the modifier key state?

There isn't one, so currently desktop on all platform is sending Android. See line 382 in this file, for instance.

krisgiesing:
I see, thanks for the clarification. I think we'll want to make one ourselves then, at some point.

stuartmorgan:
Yes, the current system of using system channels and internal implementation details should be replaced with a real embedding API (essentially #47, but for key events rather than text input).

PR #244, opened 2019-01-22
[linux] Improve example Makefile

stuartmorgan:
- Clearly separates out and documents the variables that are most likely
  to need to change if copying the example app.
- Better differentiate between paths that are relative to the example
  and paths that are relative to the FDE repository, since apps using
  FDE would not be expected to be in the same directory the way the
  example is.
- Add a minimal "release" mode that enables that flag to disable Dart
  asserts, as on macOS and Windows.

### DIFF
--- a/example/linux_fde/Makefile
+++ b/example/linux_fde/Makefile
@@ -12,13 +12,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Example-specific variables.
+# To modify this Makefile for a different application, these are the values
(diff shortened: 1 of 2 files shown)


example/linux_fde/Makefile
```diff
@@ -12,13 +12,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Example-specific variables.
+# To modify this Makefile for a different application, these are the values
```
franciscojma86: This is great! 👍

PR #250, opened 2019-01-27
Add a warning about binaries to library README

stuartmorgan:
Call out the requirement that the library and the Flutter application
must be built with the same version, since some people have
started distributing binary versions apparently without being aware
of this.


stuartmorgan:
@putraxor @PotterDai FYI.

putraxor:
thank you, it looks good to me

PotterDai:
@stuartmorgan  I just posted a notice about the versioning issue at my downloading page, if there is any problem of distributing binaries please let me know! I shared it simply because I found it much simpler to check out different packages, since I don't need to build to my phone every time.

Thank you for the remind! = )

Issue #230, opened 2019-01-12
[linux/window] Fix exported API surface to be DLL/so-safe


CallumIddon:
VS builds now have a large number of export warnings since #216 landed.

Log:
```
"C:\dev\flutter-desktop-embedding\example\windows\Example Embedder.sln" (default target) (1) ->
"C:\dev\flutter-desktop-embedding\example\windows\GLFW Example.vcxproj.metaproj" (default target) (2) ->
"C:\dev\flutter-desktop-embedding\library\windows\GLFW Library.vcxproj" (default target) (3) ->
(ClCompile target) ->
  c:\dev\flutter-desktop-embedding\library\include\flutter_desktop_embedding\method_call.h(46): warning C4251: 'flutter_desktop_embedding::MethodCal
l<T>::method_name_': class 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>' needs to have dll-interface to be used by clients o
f class 'flutter_desktop_embedding::MethodCall<T>' [C:\dev\flutter-desktop-embedding\library\windows\GLFW Library.vcxproj]
  c:\dev\flutter-desktop-embedding\library\include\flutter_desktop_embedding\method_call.h(47): warning C4251: 'flutter_desktop_embedding::MethodCal
l<T>::arguments_': class 'std::unique_ptr<_Ty,std::default_delete<_Ty>>' needs to have dll-interface to be used by clients of class 'flutter_desktop
_embedding::MethodCall<T>' [C:\dev\flutter-desktop-embedding\library\windows\GLFW Library.vcxproj]
  c:\dev\flutter-desktop-embedding\library\include\flutter_desktop_embedding\engine_method_result.h(47): warning C4251: 'flutter_desktop_embedding::
internal::ReplyManager::reply_handler_': class 'std::function<void (const uint8_t *,size_t)>' needs to have dll-interface to be used by clients of c
lass 'flutter_desktop_embedding::internal::ReplyManager' [C:\dev\flutter-desktop-embedding\library\windows\GLFW Library.vcxproj]
  c:\dev\flutter-desktop-embedding\library\include\flutter_desktop_embedding\engine_method_result.h(84): warning C4251: 'flutter_desktop_embedding::
EngineMethodResult<T>::reply_manager_': class 'std::unique_ptr<flutter_desktop_embedding::internal::ReplyManager,std::default_delete<_Ty>>' needs to
[...]

PR #253, opened 2019-01-28
[linux/windows] Add FlutterWindowController

stuartmorgan:
Creates a simple C++ object as the primary interaction point for the
embedder calls. This provides a simpler API surface than embedder.h,
and folds in some common code (e.g., error logging).

This also serves to insulate clients from the embedder.h API layer,
so that future incremental changes done for #230 will cause less churn
for embedders.

### DIFF
--- a/library/include/flutter_desktop_embedding/glfw/flutter_window_controller.h
+++ b/library/include/flutter_desktop_embedding/glfw/flutter_window_controller.h
@@ -1,0 +1,94 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


library/include/flutter_desktop_embedding/glfw/flutter_window_controller.h
```diff
@@ -0,0 +1,89 @@
...
+//
+// Note: This is an early implementation (using GLFW internally) which
+// requires control of the application's event loop, and is thus useful
+// primarily for building a simple one-window shell hosting a Flutter
+// application. The final implementation and API will be very different.
+class FDE_EXPORT FlutterWindowController {
+ public:
+  // There must be only one instance of this class in an application at any
+  // given time, as Flutter does not support multiple engines in one process,
+  // or multiple views in one engine.
```
stuartmorgan: The reason this is just a comment rather an actual enforced singleton design is that my hope is that the C++ wrapper layer I'm working on extracting for #230 (which this would be part of) could be header-only, which wouldn't work with a singleton, so I'm structuring it this way to allow flipping to header-only without more API churn.

PR #254, opened 2019-01-29
Remove GLFW from the API surface

stuartmorgan:
Rather than exposing the GLFWwindow that is used, treat it as an
implementation detail and use an opaque pointer specific to this library
as the context object instead.

In practice the use of GLFW can't be entirely internal since the use of
GLFW in the library prevents its use by the embedder, but the details of
how it is used should not be exposed. This also gives greater control
over the API surface, as part of working toward a stable ABI (#230).

This does mean that the embedder cannot manipulate the GLFW window
directly, but rather than trust that doing so would be safe across DLL
boundaries, any functionality need by the simple embedder
use cases this library enables should be exposed via wrappers.

### DIFF
--- a/library/common/glfw/embedder.cc
+++ b/library/common/glfw/embedder.cc
@@ -20,6 +20,13 @@
 #include <cstdlib>
 #include <iostream>
 
+#ifdef __linux__
+// Epoxy must be included before any graphics-related code.
+#include <epoxy/gl.h>
+#endif
+
+#include <GLFW/glfw3.h>
+
 #include <flutter_embedder.h>
 
 #include "library/common/glfw/key_event_handler.h"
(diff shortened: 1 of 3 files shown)


library/common/glfw/embedder.cc
```diff
@@ -300,62 +313,69 @@ bool FlutterInit() {
...
   // aligning more closely with the Flutter registrar system.
-  return state->plugin_handler.get();
+  return flutter_window->plugin_handler.get();
 }
 
-GLFWwindow *CreateFlutterWindow(size_t initial_width, size_t initial_height,
-                                const std::string &assets_path,
-                                const std::string &icu_data_path,
-                                const std::vector<std::string> &arguments) {
+FlutterWindowRef CreateFlutterWindow(
```
RobertZenz: Was that whitespace change deliberately?

stuartmorgan: I'm not sure what whitespace change you mean, but since we use clang-format the answer is almost certainly yes.

PR #257, opened 2019-01-30
[macos] Adjust for embedding API type name change

stuartmorgan:
https://github.com/flutter/engine/pull/7567 renamed the type of the
result for embedding API calls. Rather than force a Flutter upgrade by
using the new type name, temporarily use 'auto' for the type (with a
TODO to change it later, since auto is arguably less readable here).

This requires converting the file to ObjC++, which required a few
explicit casts in places that were previously being silently recast.


stuartmorgan:
I thought I would need this for the GLFW embedder as well, but it turns out that's already using `auto` for this type.

PR #258, opened 2019-01-30
Updated requirements with necessary import for "Platform"

ynsgnr:


ynsgnr:
Understanable, thank you for response

PR #259, opened 2019-01-31
[linux/windows] Separate out a C channel API

stuartmorgan:
Eliminates PluginRegistrar and BinaryMessenger from the embedder.h API
surface, moving them up to the FlutterWindowController level. In their
place, adds new low-level messaging APIs, which are based closely on
the Flutter engine embedder APIs (including an equivalent message
struct), in some cases wrapping them directly. Unlike the engine APIs,
they allow per-channel callbacks, so individual plugins can use the API.

Since PluginHandler is no longer the source of truth for callback
registration, only an adaptor, it's now fine to have multiple instances
of them, all passing through to the underlying C API.

Part of #230

### DIFF
--- a/library/BUILD.gn
+++ b/library/BUILD.gn
@@ -38,6 +38,8 @@ published_shared_library("flutter_embedder") {
   if (is_linux || is_win) {
     sources += [
       "common/engine_method_result.cc",
+      "common/internal/incoming_message_dispatcher.cc",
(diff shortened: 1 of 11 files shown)


stuartmorgan:
I'll need to add the new file to Windows, which I'll do tomorrow, but that won't affect the code (unless there are Windows-specific build issues, but I'm not expecting that) so it can be reviewed before then.

PR #262, opened 2019-02-01
[windows] Switch to GN for library build

stuartmorgan:
Eliminates the VS-based embedder library build and updates the
example to use the GN build.

As part of the change, moves build output to
example/build/windows_fde
to match the recent change for Linux. Also eliminates the
'Dynamic Library' portion of the configuration names since
they are relics of a distinction that was removed in an earlier
change.


stuartmorgan:
Creating the PR to test the AppVeyor changes; not requesting review because it's not ready--Debug works, but Release crashes immediately. I strongly suspect this is due to #230, since the GN build isn't using the same build flags that the Release VS build uses. I'll revisit this once that work is further along.

(Also, the READMEs need to be updated is part of this PR, which isn't done yet.)

stuartmorgan:
AppVeyor fixed, Release works now that #230 is fixed, and the READMEs are updated. Ready for review!

PR #263, opened 2019-02-01
[macOS] Add FLETexture and FLETextureRegistrar.

cloudwebrtc:
I compiled this patch using Flutter Channel dev and may need to wait for the built-in Texture support in the FlutterEmbedder.framework stable version to review and merge.


stuartmorgan:
The macOS code has already been migrated to flutter/engine, and the copy here will be removed once a prebuilt framework is available to transition clients to (per the plan outlined in my recent email to the mailing list), hopefully early next week. Given that, the copy here is frozen, and you'll need to do this as a PR in flutter/engine instead.

It should be straightforward to move over, since the code there hasn't changed other than what was done in the migration (e.g., eliminating duplication with existing iOS code).

cloudwebrtc:
This is good news, I'll try to resubmit the PR to flutter/engine. :)

Issue #265, opened 2019-02-03
Can't build using Visual Studio Community 2017


krisu7:
I can't run example app on windows 10.
I'm getting errors:
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	MSB3073	The command "C:\Users\user\Desktop\Project\dashboard\desktop\flutter-desktop-embedding\library\windows\scripts\update_flutter_engine && C:\Users\user\Desktop\Project\dashboard\desktop\flutter-desktop-embedding\library\windows\scripts\get_engine_artifacts && C:\Users\user\Desktop\Project\dashboard\desktop\flutter-desktop-embedding\library\windows\scripts\get_GLFW && C:\Users\user\Desktop\Project\dashboard\desktop\flutter-desktop-embedding\library\windows\scripts\build_jsonlib --debug
:VCEnd" exited with code 255.	GLFW Library	C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Microsoft.CppCommon.targets	128	
```
and

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	MSB3073	The command "xcopy /y /d /q "C:\Users\user\Desktop\Project\dashboard\desktop\flutter-desktop-embedding\example\windows_fde\bin\x64\Debug Dynamic Library\GLFW Example\..\GLFW Library\flutter_engine.dll" "C:\Users\user\Desktop\Project\dashboard\desktop\flutter-desktop-embedding\example\windows_fde\bin\x64\Debug Dynamic Library\GLFW Example\"
:VCEnd" exited with code 4.	GLFW Example	C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Microsoft.CppCommon.targets	133	
```

stuartmorgan:
Do you have any other build output besides the errors? The first step that's failing there is a series of scripts chained together, so it's not clear which one is failing. Many of the script steps will print message when something fails (or in some cases on successful completion, which would give more information about progress).

krisu7:
Here is my build output:
```
1>------ Build started: Project: GLFW Library, Configuration: Debug Dynamic Library x64 ------
1>Get the flutter engine, engine artifacts, GLFW and jsoncpp
1>Running git merge-base --is-ancestor 390ded9340e529b8475fefd1afdbe59c5b8d4081 HEAD
[...]

PR #266, opened 2019-02-03
Move Windows PATH instructions

stuartmorgan:
The VS build also requires the VC build tools to be in the path (due to
running jsoncpp's build via script), so move the instructions from the
GN readme to the main readme.


stuartmorgan:
Addresses issues that came up recently on both the mailing list and #265

PR #277, opened 2019-02-13
Switch to FlutterMacOS.framework

stuartmorgan:
Removes the MacOS library code, replacing it with use of the prebuilt
FlutterMacOS.framework downloaded from the Flutter binary server.

Plugin code has been updated to use the adjusted API; FLE* versions
of many classes were removed in the migration to flutter/engine, in favor
of sharing the existing iOS versions.

What was FlutterEmbedderMac.xcodeproj exists now only as a project to
provide a shared dependency for downloading the framework, and has been
moved to plugins/common/macos/SharedFlutterFramework.xcodeproj.

In order to avoid people thinking their apps should depend on that
project, the example instead fetches its own copy of the framework to a
different location. This is slightly wasteful, but simplifies
understanding of the example.

Most of the logic of update_flutter_engine.dart has been refactored
into a helper class for sharing with the new sync_flutter_library.
There is still some duplication in the top-level scripts, but since
fetch_flutter_engine will be eliminated once the migration is complete
this isn't a significant issue.


README.md
```diff
@@ -30,6 +32,19 @@ containing flutter-desktop-embedding, containing a path to the Flutter tree to
 use, if you prefer not to have the Flutter tree next to
 flutter-desktop-embedding.
 
+#### Tools
+
+You will developer tools for your platform:
```
franciscojma86: you will need*

stuartmorgan: Fixed

PR #286, opened 2019-02-20
Add a simple dependency checking script

stuartmorgan:
This should slightly simplify the process of figuring out if people have
missed a setup step when getting started with the project.

### DIFF
--- a/tools/dart_tools/bin/doctor.dart
+++ b/tools/dart_tools/bin/doctor.dart
@@ -41,6 +41,15 @@ Future<void> main() async {
   }
   if (Platform.isLinux) {
     dependencies.add(_Dependency('GCC', 'g++', ['--version']));
+    dependencies.add(_Dependency('libraries', 'pkg-config', [
+      '--exists',
+      '--print-errors',
+      'epoxy',
+      'glfw3',
+      'gtk+-3.0',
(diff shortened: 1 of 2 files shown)


tools/dart_tools/bin/doctor.dart
```diff
@@ -0,0 +1,75 @@
...
+  final dependencies = [
+    _Dependency('git', 'git', ['--version']),
+  ];
+  if (Platform.isWindows || Platform.isLinux) {
+    dependencies
+      ..add(_Dependency('GN', 'gn', ['--version']))
+      ..add(_Dependency('ninja', 'ninja', ['--version']));
+  }
+  if (Platform.isLinux) {
+    dependencies.add(_Dependency('GCC', 'g++', ['--version']));
```
franciscojma86: What about glfw and other mentioned in the readme?

stuartmorgan: I was thinking that it would be tricky to check them in a general way, so wasn't going to worry about it for now (especially since having those missing should give understandable build errors), but realized when I was about to say so that it's actually just a simple pkg-config command. Added :)

franciscojma86: Nice, thanks!

PR #295, opened 2019-02-28
Update README.md

fmatosqg:
Add dependency to g++, tested on a newly installed mint 19.1

### DIFF
--- a/library/README.md
+++ b/library/README.md
@@ -51,12 +51,13 @@ First you will need to install the relevant library dependencies:
 *   epoxy
 *   X11 development libs
 *   pkg-config
+*   g++
 
 Installation example for debian-based systems:
 
 ```
 $ sudo apt-get install libglfw3-dev libepoxy-dev libjsoncpp-dev libgtk-3-dev \
-      libx11-dev pkg-config
+      libx11-dev pkg-config g++
 ```
 
 #### Using the Library


stuartmorgan:
This is already covered here:
https://github.com/google/flutter-desktop-embedding#tools

What's the reasoning for duplicating this? I'd rather not cover the same information in two places.

fmatosqg:
Well, the information is not the same (root readme mentions gcc vs g++). Also it lowers the threshold of knowledge/time needed to get this going. 

I could perfectly time figuring out all the dependencies from that apt-get all by myself, but it's definitely convenient to have that in a README file. This fix is about improving convenience.

You can argue having gcc or g++ installed is trivial, but that's not my point.

stuartmorgan:
> Well, the information is not the same (root readme mentions gcc vs g++).

That's just because I'm used to both as being generally referred to as gcc, even though that's not technically accurate. I'm fine with changing the root README to say g++ if you found that mismatch confusing.

> You can argue having gcc or g++ installed is trivial, but that's not my point.

I don't think whether it's trivial or not is relevant; what does matter is the audience that these READMEs are targeting. Specifically, as noted on that page "Currently the development workflow assumes you are starting with an existing desktop application project". We don't provide step-by-step instruction on installing Xcode or Visual Studio either.

Fundamentally, if someone doesn't either already have (or know how to install) the relevant compilation tools, they are not the target audience for the project in its current state.

fmatosqg:
Ok let's drop the pr, it's no big deal.

 I think that our opinions differ on something more fundamental. The Linux
set-up was so straightforward that you're probably scaring off people who
just wants to have some fun with dart and Flutter and don't need native
plugins.

Yep I did have problems in my Ubuntu related to library conflicts, but a
fresh Linux install or a docker image make things light years easier than
os x or VStudio. And electron apps show how thirsty people are for simple
apps.

stuartmorgan:
> you're probably scaring off people who just wants to have some fun with dart and Flutter

Someone who wants to have fun with Flutter and is scared off by needing to install g++ without a command to do it would be much better off using Flutter itself, which is supported, well documented, and focused on ease of use—not an experimental, unstable project that requires manually building applications from source.

> a fresh Linux install or a docker image make things light years easier

Anyone who wants to write and maintain instructions somewhere on how to set up and use this project from scratch using docker images is welcome to do so; that's not a focus for this project at the moment.

> And electron apps show how thirsty people are for simple apps.

Presumably people building and distributing Electron apps want a production-ready stack, which this project isn't yet. Making the setup instructions slightly easier isn't going to change the fact that, e.g., scrolling and right-click don't work. Our focus is on addressing the latter.

PR #318, opened 2019-04-03
Use the prebuilt shells on Windows and Linux

stuartmorgan:
- Removes the library/ directory, since all that code has migrated to flutter/engine.
- Removes some tooling that is no longer used (e.g., update_flutter_engine, fetch_glfw).
- Updates the Windows and Linux examples to download the library directly, rather than
   rely on a GN-downloaded copy, matching macOS.
- Moves the GN rules to download a library for use by the plugins to plugins/common,
   matching macOS.
- Does a first pass of updating the READMEs to reflect the fact that the shells are no
   longer conceptually part of this project (including pointing to the new Flutter
   desktop wiki page).
- Moves some information that was in library/README.md and is still useful, but not
   moved to the Flutter wiki, to other READMEs.

Note: Although the Windows example now doesn't actually rely on anything built by
GN, the include paths and the call to run the GN build remain to avoid the churn of
removing them now and then re-adding them when plugins are added.


stuartmorgan:
This ended up being somewhat of a mega-patch; originally I was going to move Linux first, then Windows in another CL, then clean everything up in a third, but ultimately decided that the intermediate states were weird and it was clearer to just do one clean switch-over.

stuartmorgan:
James, adding you as a potential reviewer since you're already working with this PR and thus familiar with parts of it already. No pressure though if a full review would be a distraction from your Windows work though.

clarkezone:
Thx Stuart, sounds good

On Apr 6, 2019, at 13:33, stuartmorgan <notifications@github.com<mailto:notifications@github.com>> wrote:


James, adding you as a potential reviewer since you're already working with this PR and thus familiar with parts of it already. No pressure though if a full review would be a distraction from your Windows work though.

stuartmorgan:
Thanks for the review!

> Looks fine modulo the issue with copying local, not sure if you want to treat that as a separate issue

Let's follow up on that post-submission, since we're not sure how common it is, and landing this even if it's not yet 100% correct for the development case is still an improvement over current situation of building source that's increasingly out of date instead of using the new framework at all.

PR #327, opened 2019-04-17
Switch plugins to standard codec

stuartmorgan:
Removes all use of the JSON codec, as well as docs and dependency
management related to jsoncpp, which is no longer a dependency.


stuartmorgan:
@krisgiesing Ping?

krisgiesing:
BTW, sorry for the late review. I never saw the github notification somehow.

PR #336, opened 2019-04-19
Refactor project to remove need for build.sh

jonahwilliams:
Removes indirection by moving the macOS build script to flutter_tools.  This still uses the FDE asset syncing. This uses almost the same scheme/flag passing setup as the existing iOS build. 

~~Currently the track-widget-creation flag is still broken though, it seems like some part of the process is  using the regular .dill instead of the .track.dill - still investigating.~~

Fixed, need to make sure I referenced config file

See also: https://github.com/flutter/flutter/pull/31329


example/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -366,7 +367,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "\"$FDE_ROOT\"/tools/build_flutter_assets \"$PROJECT_DIR\"/..\n";
+			shellScript = "\"$FLUTTER_ROOT\"/packages/flutter_tools/bin/macos_build_flutter_assets.sh";
```
stuartmorgan: This creates a new requirement that `flutter build` must be run before the Xcode project will work; that's not an issue for a long-term `flutter create`d version of this, but would be potentially breaking for this project. I think we should check in a Generated.xcconfig with this change that defaults to `../../../flutter` to make it work in a clean checkout that uses the suggested layout. (Not perfect, since it won't handle .flutter_location_config, but better than nothing.)

example/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -447,6 +448,7 @@
 /* Begin XCBuildConfiguration section */
 		33CC10F92044A3C60003C045 /* Debug */ = {
 			isa = XCBuildConfiguration;
+			baseConfigurationReference = 7AFA3C8E1D35360C0083082E /* Generated.xcconfig */;
```
stuartmorgan: We should follow the iOS model of having a Debug.xcconfig and a Release.xcconfig that are used as bases, and have them import the Generated.xcconfig. That makes things more consistent between the two platforms, and gives us flexibility in the future.

jonahwilliams: Done

example/macos/Runner.xcodeproj/project.pbxproj
jonahwilliams: Putting it together this way it doesn't quite work when i run from Xcode - possibly due to the lack of the OBJROOT, SYMROOT, and derivedData overrides?  I might have my setup configured wrong.

stuartmorgan: I'll take a look tomorrow and see what's broken. The build output will go somewhere else if you build from Xcode, but I wouldn't expect that to break anything (it's what happens right now with build.sh vs Xcode UI).

stuartmorgan: I realized while doing the Windows version that we shouldn't check in a copy of the generated settings, since it needs to be in the .gitignore, so I think we'll just live with needing to use 'fluter build' or 'flutter run' at least once. I can add a note about that to the docs.

example/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -148,6 +150,15 @@
 			name = Frameworks;
 			sourceTree = "<group>";
 		};
+		D76E93BA226BCA8F00F7932B /* Recovered References */ = {
```
stuartmorgan: The configs should be referenced in the project, rather than being considered recovered files. Please rename this group in the file tree "Configs", move it up between Resources and Frameworks, and (for ease of inspection) add Generated.xcconfig to that group as well (make sure to uncheck the option to add it to any of the targets)

jonahwilliams: We don't reference the Generated.xcconfig in our iOS projects, should this be changed in the templates for those as well?

stuartmorgan: It doesn't affect the build in any way, it just means that if you have the project open you can easily inspect all the configs. I find it helpful in general, but if nobody has ever complained it's up to you whether it's worth changing iOS.

jonahwilliams:
Looks like travis will be red until it uses the latest flutter

stuartmorgan:
> Looks like travis will be red until it uses the latest flutter

It clones from master; it's red because your other PR hasn't landed...

example/.gitignore
```diff
@@ -13,3 +13,6 @@ packages
 # Ignore it since it just creates noise. When there's a system in place
 # for versioning this project relative to Flutter, remove this.
 pubspec.lock
+
+# Generated by flutter run and flutter build
+Generated.xcconfig
```
stuartmorgan: This should be in a .gitignore in example/macos (especially since the goal is to make those folders relocatable shortly)

testbed/.gitignore
```diff
@@ -13,3 +13,6 @@ packages
 # Ignore it since it just creates noise. When there's a system in place
 # for versioning this project relative to Flutter, remove this.
 pubspec.lock
+
+# generated by flutter run/build
+Generated.xcconfig
```
stuartmorgan: Same.

testbed/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -366,7 +378,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "\"$FDE_ROOT\"/tools/build_flutter_assets \"$PROJECT_DIR\"/..\n";
+			shellScript = "\"$FLUTTER_ROOT\"/packages/flutter_tools/bin/macos_build_flutter_assets.sh";
```
stuartmorgan: Don't change this yet for testbed; that would break the documented (and heavily used by me) local engine workflow.

Later we can make the Flutter scripts support local engine and migrate the workflow and docs to that.

PR #346, opened 2019-04-22
Support --track-widget-creation on Windows

stuartmorgan:
Plumbs the flag for enabling widget creation tracking from the build
script through to 'flutter build bundle'. Fixes the crash-on-hot-reload
in VS Code for Windows.

Since this added the logic to generate a property file for VS with
settings, also plumbs through FLUTTER_ROOT, although doesn't yet rely on
it being there in the example.

### DIFF
--- a/example/windows/.gitignore
+++ b/example/windows/.gitignore
@@ -3,6 +3,9 @@
 ##
 ## Get latest from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore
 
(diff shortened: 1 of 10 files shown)


stuartmorgan:
Obviously we'll want to move the property file generation logic into flutter_tools at some point rather than have duplicate copies in each project.

This requires that 'flutter build' or 'flutter run' be used once before building from within VS will work.

PR #348, opened 2019-04-22
Make example/windows/ relocatable

stuartmorgan:
Removes dependencies from example/windows/ on the FDE project, replacing the
'build bundle' wrapper with a direct call, and the flutter library sync with a manual
copy.

This allows windows/example/ to be copied into an arbitrary Flutter project to
enable build/run for Windows. The dart code would of course still need changes
as described at:
https://github.com/flutter/flutter/wiki/Desktop-shells#flutter-application-requirements


stuartmorgan:
The overall functionality of the scripts should definitely move into the Flutter folder as a future step. I think that will require more thought though, since we should figure out what exactly that should look like (e.g., I strongly suspect we should use Dart, not Batch, for the version of this stuff that lives in flutter_tool).

Doing this first let me make a small local change that unlocks people being able to copy the directory easily, and then we can iterate on the real solution later.

PR #349, opened 2019-04-22
Add macos testbed TRACK_WIDGET_CREATION handling

stuartmorgan:
Testbed doesn't yet use the Flutter script for macOS bundle build, to
preserve the local engine build workflow, so handle the build setting
manually in the bundle script.


stuartmorgan:
In combination with your patch to update the macOS builds in general, this will fix the track widget issues for testbed/.

PR #351, opened 2019-04-23
Make example/linux/ relocatable

stuartmorgan:
Removes dependencies from example/linux/ on the FDE project, replacing the
flutter library sync with a manual copy and the bundle build with a
direct call to the flutter tool.

Since make doesn't have a standard format for out-of-file settings, for
now this just uses a dot file to store the Flutter root, which is the
only critical setting to preserve at this stage. Longer term, a more
robust solution for persisting arguments will likely be needed.

This allows example/linux/ to be copied into an arbitrary Flutter project to
enable build/run for Linux. The dart code would of course still need changes
as described at:
https://github.com/flutter/flutter/wiki/Desktop-shells#flutter-application-requirements


stuartmorgan:
Obviously as with the other platforms, the logic for copying artifacts to a project-specific location should at some point move into flutter_tool

PR #353, opened 2019-04-23
Remove need for separate asset script

jonahwilliams:
Moves framework sync and local engine flags into tool in https://github.com/flutter/flutter/pull/31515

### DIFF
--- a/example/macos/Runner.xcodeproj/project.pbxproj
+++ b/example/macos/Runner.xcodeproj/project.pbxproj
@@ -19,17 +19,6 @@
 			name = "Build Flutter Bundle";
 			productName = FLX;
 		};
-		33D1A0E422148440006C7A3E /* Copy Flutter Framework */ = {
-			isa = PBXAggregateTarget;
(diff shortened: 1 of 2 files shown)


jonahwilliams:
Not sure if I should remove this for testbed as well...

stuartmorgan:
> Not sure if I should remove this for testbed as well...

I was going to do some cleanup on testbed/ today, and as part of that I'll evaluate how to handle the transition in a way that won't make the local development workflow overly complex in the short term. (The current setup evolved precisely to avoid the problem of having to change several things in several places to get local development working, so I want to avoid having things split too much as the tooling is evolving.)

stuartmorgan:
Looking again at the way engine override is plumbed: with this change we should be able to switch testbed over to using the same flow as example/, since the script handles both `build bundle` and the asset copying, which is all the the old flow was managing. \o/

stuartmorgan:
`testbed` is broken because it was still using `build_flutter_assets` instead of calling the `flutter_tools` script (since that didn't used to have local engine support). You'll need to update that script step to match the earlier `example` change to use the shared script.

PR #360, opened 2019-04-29
Add a min Flutter version to the example

stuartmorgan:
This will help catch some errors caused by version mismatches that would
otherwise cause error messages that aren't at all self-explanatory.
This won't catch all cases, but for those that it does the failure will
be very early, and much clearer.

The intent is to rev this number every time a change is made on the
Flutter side that requires a corresponding change in this repository.

### DIFF
--- a/example/pubspec.yaml
+++ b/example/pubspec.yaml
@@ -1,6 +1,13 @@
 name: example_flutter
 description: An example project for flutter-desktop-embedding.
 
+environment:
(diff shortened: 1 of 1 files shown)


stuartmorgan:
I've not added this to `testbed` intentionally, since having to update it in two places will be more overhead, and I expect far fewer people to be building `testbed`, and for them to be much more likely to know they need to be closely tracking Flutter master (and to understand why something is failing if they forget to update Flutter when updating FDE).

PR #365, opened 2019-05-03
Add dark theme to example

jonahwilliams:
For a nice surprise on macOS


stuartmorgan:
> what about an auto detection feature of dark mode?

That's what https://github.com/flutter/engine/pull/8810 added; this PR would cause the example app to opt in to changing appearance based on that auto-detection.

> Since osx can have that setup as default, it might make sense to prepare an app automatically to match the os dark/light theme.

Providing a non-null `darkTheme`, as this proposed PR does for the example app, is exactly how someone would do that. See
https://docs.flutter.io/flutter/material/MaterialApp/darkTheme.html

There are lots of things that might make sense for an app to do, but demonstrating all of them is explicitly not the goal of `example`, per my explanation above.

jonahwilliams:
@sassman to be clear, auto detection of light/dark mode is already supported on Android and now macOS. The developer is required to supply a dark theme to the material app to transition between them

stuartmorgan:
What about adding this to `testbed` per my earlier comment? I think it would be perfect to add there.

jonahwilliams:
Ahh I completely missed that, will re-open when I have a few minutes

stuartmorgan:
Updated to add the line to `testbed` instead.

PR #367, opened 2019-05-06
Implement window resize plugin for Linux

### LINKED ISSUE #99: Allow window size/positioning control
It would be useful for the Flutter application to be able to easily control the window size and position. To be useful, this will also require the ability to query the monitor size(s) (and positions for multiple monitors).

This can be done pretty simply with a plugin.

stuartmorgan:
Adds a Linux implementation of the window resize plugin using the new
APIs in the library.

Fixes #99

### DIFF
--- a/plugins/window_size/linux/include/window_size/window_size_plugin.h
+++ b/plugins/window_size/linux/include/window_size/window_size_plugin.h
@@ -1,0 +1,38 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#ifndef PLUGINS_WINDOW_SIZE_LINUX_INCLUDE_WINDOW_SIZE_WINDOW_SIZE_PLUGIN_H_
+#define PLUGINS_WINDOW_SIZE_LINUX_INCLUDE_WINDOW_SIZE_WINDOW_SIZE_PLUGIN_H_
+
+// A plugin to allow resizing the window.
+
+#include <flutter_plugin_registrar.h>
+
+#ifdef WINDOW_SIZE_PLUGIN_IMPL
+#define WINDOW_SIZE_PLUGIN_EXPORT __attribute__((visibility("default")))
+#else
+#define WINDOW_SIZE_PLUGIN_EXPORT
+#endif
+
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+WINDOW_SIZE_PLUGIN_EXPORT void WindowSizeRegisterWithRegistrar(
+    FlutterDesktopPluginRegistrarRef registrar);
+
+#if defined(__cplusplus)
+}  // extern "C"
+#endif
+
+#endif  // PLUGINS_WINDOW_SIZE_LINUX_INCLUDE_WINDOW_SIZE_WINDOW_SIZE_PLUGIN_H_
--- a/plugins/window_size/linux/src/window_size_plugin.cc
+++ b/plugins/window_size/linux/src/window_size_plugin.cc
@@ -195,6 +195,8 @@ void WindowSizePlugin::HandleMethodCall(
       result->Error("Bad arguments", "Expected 4-element list");
       return;
     }
+    // Frame validity (e.g., non-zero size) is assumed to be checked on the Dart
+    // side of the call.
     const auto &frame_list = method_call.arguments()->ListValue();
     flutter::WindowFrame frame = {};
     frame.left = static_cast<int>(frame_list[0].DoubleValue());
--- a/plugins/window_size/linux/src/window_size_plugin.cc
+++ b/plugins/window_size/linux/src/window_size_plugin.cc
@@ -1,0 +1,221 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#include "plugins/window_size/linux/include/window_size/window_size_plugin.h"
+
+#include <gtk/gtk.h>
+#include <iostream>
+#include <memory>
+#include <vector>
+
+#include <flutter/flutter_window.h>
+#include <flutter/method_channel.h>
+#include <flutter/plugin_registrar_glfw.h>
+#include <flutter/standard_method_codec.h>
+
+#include "plugins/window_size/common/channel_constants.h"
+
+namespace plugins_window_size {
+
+namespace {
+
+using flutter::EncodableList;
+using flutter::EncodableMap;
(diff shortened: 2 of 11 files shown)


stuartmorgan:
This will need a slight update to set a newer min Flutter version in the update script, but I need to wait for the next Flutter engine roll to have a hash for that; sending for review in parallel since that won't be an interesting change to review.

testbed/lib/main.dart
```diff
@@ -31,13 +31,13 @@ void main() {
 
   // Try to resize and reposition the window to be half the width and height
   // of its screen, centered horizontally and shifted up from center.
-  if (Platform.isMacOS) {
+  if (Platform.isMacOS || Platform.isLinux) {
     window_size.getWindowInfo().then((window) {
       if (window.screen != null) {
         final screenFrame = window.screen.visibleFrame;
-        final width = math.max((screenFrame.width / 2).roundToDouble(), 640.0);
+        final width = math.max((screenFrame.width / 2).roundToDouble(), 800.0);
```
krisgiesing: This change sees unrelated to the rest of the PR, is that true?

plugins/window_size/linux/src/window_size_plugin.cc
```diff
@@ -0,0 +1,221 @@
...
+
+namespace plugins_window_size {
+
+namespace {
+
+using flutter::EncodableList;
+using flutter::EncodableMap;
+using flutter::EncodableValue;
+
+// Returns the serialiable form of |frame| expected by the platform channel.
```
krisgiesing: Nit: typo "serializable"

plugins/window_size/linux/src/window_size_plugin.cc
```diff
@@ -0,0 +1,221 @@
...
+    const GdkRectangle &frame) {
+  return EncodableValue(EncodableList{
+      EncodableValue(static_cast<double>(frame.x)),
+      EncodableValue(static_cast<double>(frame.y)),
+      EncodableValue(static_cast<double>(frame.width)),
+      EncodableValue(static_cast<double>(frame.height)),
+  });
+}
+
+// Extracts information from |monitor| and returns the serialiable form expected
```
krisgiesing: Nit: typo "serializable"

plugins/window_size/linux/include/window_size/window_size_plugin.h
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#ifndef PLUGINS_WINDOW_SIZE_LINUX_INCLUDE_WINDOW_SIZE_WINDOW_SIZE_PLUGIN_H_
+#define PLUGINS_WINDOW_SIZE_LINUX_INCLUDE_WINDOW_SIZE_WINDOW_SIZE_PLUGIN_H_
+
+// A plugin to allow resizing the window.
+
+#include <flutter_plugin_registrar.h>
+
+#ifdef WINDOW_SIZE_PLUGIN_IMPL
```
krisgiesing: Is this idiom used across all the plugins? It might be nice documenting it somewhere. It's not immediately obvious at this site where WINDOW_SIZE_PLUGIN_IMPL is expected to be defined.

plugins/window_size/linux/src/window_size_plugin.cc
```diff
@@ -0,0 +1,221 @@
...
+          (frame.y >= monitor_frame.y &&
+           frame.y <= monitor_frame.y + monitor_frame.width)) {
+        return monitor;
+      }
+    }
+  }
+  return nullptr;
+}
+
+// Extracts information from |window| and returns the serialiable form expected
```
krisgiesing: Nit: typo "serializable"

plugins/window_size/linux/src/window_size_plugin.cc
```diff
@@ -0,0 +1,221 @@
...
+
+  // Uses new instead of make_unique due to private constructor.
+  std::unique_ptr<WindowSizePlugin> plugin(
+      new WindowSizePlugin(std::move(channel), registrar->window()));
+
+  channel_pointer->SetMethodCallHandler(
+      [plugin_pointer = plugin.get()](const auto &call, auto result) {
+        plugin_pointer->HandleMethodCall(call, std::move(result));
+      });
+  registrar->EnableInputBlockingForChannel(kChannelName);
```
krisgiesing: Maybe add a comment on what would happen if input blocking wasn't enabled?

plugins/window_size/linux/src/window_size_plugin.cc
```diff
@@ -0,0 +1,221 @@
...
+    }
+    result->Success(&screens);
+  } else if (method_call.method_name().compare(kGetWindowInfoMethod) == 0) {
+    EncodableValue window_info =
+        GetPlatformChannelRepresentationForWindow(window_);
+    result->Success(&window_info);
+  } else if (method_call.method_name().compare(kSetWindowFrameMethod) == 0) {
+    if (!method_call.arguments() || !method_call.arguments()->IsList() ||
+        method_call.arguments()->ListValue().size() != 4) {
+      result->Error("Bad arguments", "Expected 4-element list");
```
krisgiesing: I'm not entirely sure but this looks like it would show as an error to developers who invoked the plugin incorrectly. Maybe report what the four elements in the list mean (e.g. doubles in TLWH format)?

plugins/window_size/linux/src/window_size_plugin.cc
```diff
@@ -0,0 +1,221 @@
...
+      return;
+    }
+    const auto &frame_list = method_call.arguments()->ListValue();
```
krisgiesing: Should we sanity check the arguments here? E.g. negative width/height disallowed

plugins/window_size/linux/src/window_size_plugin.cc
```diff
@@ -0,0 +1,221 @@
...
+      GdkMonitor *monitor = gdk_display_get_monitor(display, i);
+      screens.ListValue().push_back(
+          GetPlatformChannelRepresentationForMonitor(monitor));
+    }
+    result->Success(&screens);
+  } else if (method_call.method_name().compare(kGetWindowInfoMethod) == 0) {
+    EncodableValue window_info =
+        GetPlatformChannelRepresentationForWindow(window_);
+    result->Success(&window_info);
+  } else if (method_call.method_name().compare(kSetWindowFrameMethod) == 0) {
```
krisgiesing: Out of scope for this PR but it might be a nice ergonomic enhancement if there were separate channel calls to set position without changing size, and size without changing position.

testbed/lib/main.dart
stuartmorgan: Not exactly. When I wrote this code on macOS to demo/exercise the new plugin, I forgot that `testbed` was 800x600, not 640x480. While testing it on Linux on a small screen with low resolution, I realized that this would actually resize testbed to be smaller on some configurations, which wasn't intended.

krisgiesing: Makes sense. Thanks! Resolving.

plugins/window_size/linux/src/window_size_plugin.cc
krisgiesing: Ah. In that case, I'd suggest documenting the assumption made here that the sanity checking is already done (otherwise it might be tempting for some future reader to "fix" things by adding code here).

plugins/window_size/linux/src/window_size_plugin.cc
krisgiesing: Ah. In that case, I'd suggest documenting the assumption made here that the sanity checking is already done (otherwise it might be tempting for some future reader to "fix" things by adding code here).

plugins/window_size/linux/src/window_size_plugin.cc
krisgiesing: Got it, thanks.

plugins/window_size/linux/src/window_size_plugin.cc
krisgiesing: SGTM

plugins/window_size/linux/src/window_size_plugin.cc
stuartmorgan: Done.

stuartmorgan:
Updated to use pre-GTK-3.22 APIs, since Xenial (which the CI uses) has 3.18. It ended up being pretty mechanical, luckily.

PR #369, opened 2019-05-08
Modified VSCode Variable to new one since old one is not working adde…

canewsin:
…d a simple linux bash script to enable desktop mode and run flutter

also added
.vscode
.flutter_location_config
to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 out/
 .DS_Store
+.vscode
+.flutter_location_config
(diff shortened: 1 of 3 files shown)


stuartmorgan:
Please don't lump unrelated changes into a single PR.

I've fixed the variable mistake in the instructions; thanks for flagging that.

As for the other changes:
- `.flutter_location_config` goes outside of the repository, by design, so putting it in `.gitignore` serves no purpose.
- If someone is using VS Code they should really put `.vscode` in a user-level git ignore, rather than expecting every project to do it.
- For a number of reasons, I prefer people to see the specifics of how to run, including the fact that it's behind an environment variable. If someone wants to streamline that locally once they understand what's happening, there are many ways to do that trivially (including ones that would work in other folders, such as an alias).

PR #373, opened 2019-05-11
Update for macOS build script changes

stuartmorgan:
Updates example and testbed for the changes in
https://github.com/flutter/flutter/pull/32538

Also moves the non-generated xcconfigs out of Flutter/; this is
inconsistent with iOS, but creates a clean separation where all of
Flutter/ is generated.


stuartmorgan:
I'll try to remember to update the min SDK version in this patch once the Flutter patch lands (so I have a target version)

stuartmorgan:
I missed that this wasn't approved at the same time as 32538, and landed that; since trunk would be broken without this, I'm going to land it TBR. (The changes are mechanical, so review is likely to be rubber-stamp anyway.)

Issue #357, opened 2019-04-27
Create a trivial example plugin


stuartmorgan:
It would be useful to have a trivial plugin example that can be used to prototype a plugin template for `flutter create`, as `example` does for the app template. Like `example`, it would have no dependencies on FDE (e.g., no GN).

This will require at least a preliminary model for what building plugin looks like on each platform (e.g., does the including app use source, or link a binary), although it may not end up being the final approach.

PR #393, opened 2019-05-20
Add an example plugin for Linux

stuartmorgan:
Adds a trivial example plugin (currently Linux-only, but will support all platforms)
to serve as a starting point for new plugins, and a place to iterate on the
plugin build/management flow.

Linux part of #357

### DIFF
--- a/plugins/example_plugin/linux/Makefile
+++ b/plugins/example_plugin/linux/Makefile
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Example-specific variables.
-# To modify this Makefile for a different application, these are the values
-# that are mostly likely to need to be changed.
+# ====================
+# Core variables. To modify this Makefile for a different plugin, these are
+# the values that are mostly likely to need to be changed.
 
 # The name of the plugin.
+# The primary source file is assumed to be $(PLUGIN_NAME).cc, with
+# $(PLUGIN_NAME).h as the public header meant for inclusion by the application.
 PLUGIN_NAME=example_plugin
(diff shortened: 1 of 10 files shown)


plugins/example_plugin/linux/Makefile
```diff
@@ -0,0 +1,139 @@
...
+
+# If FLUTTER_ROOT is set, use that version of Flutter. Otherwise, assume that
+# the flutter tool is in the path.
+ifeq ($(strip $(FLUTTER_ROOT)),)
+FLUTTER_BIN=flutter
+else
+FLUTTER_BIN=$(FLUTTER_ROOT)/bin/flutter
+endif
+
+# Unpack configuration.
```
jonahwilliams: Currently working on a unified build API at go/flutter-assemble

plugins/example_plugin/linux/Makefile
```diff
@@ -0,0 +1,139 @@
+# Copyright 2019 Google LLC
```
jonahwilliams: Is the idea that each plugin essentially has its own Makefile? Would this be something generated by the tool or provided by the developer?

stuartmorgan: This is very much not yet what I expect plugin builds to look like, it's essentially just a Make version of the current (GN) approach. E..g, the macOS version of this for now will have an Xcode project just like all the other plugins currently do, but longer term I don't think that's how they will actually work (since I think we'll follow the iOS model).

If we do have to keep something like this for Linux we'd want it to be part of the `create` template, but that's still TBD.

(I'll retool the plugin README at some point to be clearer about the extent to which example_plugin's build is preliminary, rather than a long-term solution.)

PR #394, opened 2019-05-20
Migrate Linux plugins to make, and remove GN

### LINKED ISSUE #392: Remove GN dependency
Originally GN was introduced to manage the need to build the framework itself across three platforms, and the issues with managing native projects given the frequent changes in files. That code has since all moved to flutter/engine (where it is still using GN), leaving GN used only for plugins.

Given that plugins very rarely require project changes, the overhead of per-platform builds is minimal; certainly not enough to justify the work of adding a macOS toolchain. In addition, FDE is deliberately moving more toward how things will be structured once Flutter desktop support is more mature, and that's very unlikely to involve building plugins with GN.

In practice only Linux is currently using GN (Windows `testbed` required it as a placeholder for future plugin support), so this just involves building working Makefiles for the plugins. I have that essentially done locally since once there was an example plugin Makefile it was straightforward, so this should happen in the next few days.

stuartmorgan:
Updates the existing Linux plugins to use 'make', using slightly modified
copies of the example_plugin Makefile. As part of that change, restructures
the Linux plugin sources to remove the src/ and include/... structure since
it was a legacy of older structures (C++ namespaces expressed via folder
name, direct inclusion of headers from the tree rather than the build output).

This moves the plugins a small step closer to what actual Linux plugin
development will look like (since to the extent that there is any build
logic in plugins themselves on Linux, the template would almost certainly
use Makefiles rather than GN).

This removes the last use of GN, so all the GN infrastructure is removed
from the project as well.

Fixes #392


stuartmorgan:
@jonahwilliams Ping

PR #406, opened 2019-05-28
Remove disabling of asserts from example runners

stuartmorgan:
Make it more clear to people using the example little or no context that
there's currently no actual support for a release mode.


stuartmorgan:
With `flutter` tool support, and more people "repackaging" the instructions for getting started without including caveats or in some cases even links to the official documentation, it would be easy for someone to think that building with the `--release` flag would give them an actual release build, which isn't the case.

PR #408, opened 2019-05-29
Implement example_plugin for Windows

### LINKED ISSUE #357: Create a trivial example plugin

stuartmorgan:
Adds an example_plugin implementation for Windows.

As with other platforms, the current build approach is a placeholder
which will be replaced with real dependency management in the future.
For now, the structure is that the plugin project is designed to only be
buildable from a Runner.sln; it uses Generated.props from the
Runner's source directory, and it builds into the Runner's build
directory.

As with the other platforms, this does not have dependencies on the rest
of FDE, should should be copyable as a starting point for making other
experimental plugins on Windows.

Also updates the plugin README to reflect that Windows now has an
example, and does some minor cleanup (e.g., standardizing example
code on using the example plugin).

Fixes #357

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -11,7 +11,11 @@ from this repository, and you must manually manage the linking and registration
 of plugins in your application (unlike on mobile, where the `flutter` tool
 handles that automatically).
 
-### Flutter
+An overview of the approach for each platform is below. See the `testbed`
+example of including optional plugins, including the changes to each platform's
+runner in the corresponding platform directory.
+
+### Dart
 
 Add local package references for the plugins you want to use to your
 pubspec.yaml. For example:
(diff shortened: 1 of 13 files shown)


stuartmorgan:
I had to make this uglier unfortunately; I forgot the which steps were done in `flutter run` before actually kicking off the build vs in the tool_backend script, and as a result clean builds didn't work. Now instead of copying the cache from the runner, it uses the runner's Generated.props and runs `unpack` manually.

PR #411, opened 2019-05-31
Update macOS projects for new plugin tooling

stuartmorgan:
First half of updating for https://github.com/flutter/flutter/pull/33636

For example:
- Switched from placeholder PluginRegistrant to empty.
  GeneratedPluginRegistrant, as with a default iOS project.

For testbed:
- Same change as to example/ as a starting point.
- Removed the direct dependency on example_plugin.
- Ran 'flutter run', which created the rest of the changes.

For example_plugin:
- Removed the Xcode project and framework-related files.
- Updated structure and file naming to match iOS plugin format.
- Added a podfile.

For all other plugins:
- Commented out macosPrefix so that the new tooling won't
  attempt to treat them as pods and fail.

The second half will be updating the other plugins to pods;
that's left as a follow-up to limit the scope of this patch.


stuartmorgan:
(The commit history is a mess since this was developed incrementally with a lot of manual steps later replaced by tool output, pausing to do things in separate PRs then merging them back in, etc. and it didn't seem worth the effort to clean up since it'll be squashed on commit anyway.)

PR #413, opened 2019-05-31
Remove 'tools' and 'third_party`

stuartmorgan:
There is no longer any project-specific tooling needed for any workflow,
so the entire tools directory can now be removed, along with the
supporting third_party Dart packages.

Updates README to remove references to it, and the instructions for
checking out Flutter in a specific location.


Nipodemos:
hello, sorry to bother, but i have a question:

does that means that it doesn't need the flutter-embedded-desktop folder to be in same folder of flutter? since you removed the instruction to do that.

stuartmorgan:
> does that means that it doesn't need the flutter-embedded-desktop folder to be in same folder of flutter?

Yes.

Issue #95, opened 2018-07-10
Support keyboard shortcuts


stuartmorgan:
The current menu support doesn't allow for keyboard shortcuts, which is definitely necessary.

(It might also be useful to have some kind of generic way for widgets to handle keyboard shortcuts, but that would need more thought about how focus works in a desktop Flutter environment.)

AppleEducate:
This is also really important for iPad development. It is now common for there to be a connected keyboard in iOS. And holding down the command key reveals all available shortcuts for an app.

stuartmorgan:
This is primarily tracking adding functionality to the menubar plugin, which is specific to this project. If you want keyboard shortcut support in Flutter itself, you'd want to file (or vote for) a request in the Flutter project's issue tracker.

AppleEducate:
Oh for sure, just thought it was still relevant to mention as there maybe some overlap.

AppleEducate:
Is there anything that could be added on macos given the MacOS Catalina release and iPadOS?

stuartmorgan:
Sorry, I don't understand the question. What impact do you believe 10.15 has on the implementation of keyboard shortcuts in macOS menubars? I don't see any changes to the `NSMenu` or `NSMenuItem` APIs.

AppleEducate:
Sorry, I meant to say does this change anything you have done for the initial menubar plugin? 

but you answered my question looks like it didn't.

stuartmorgan:
I'm still not understanding. The APIs for setting menubar shortcuts have been the same since macOS 10.0, so I'm not clear on why you would expect 10.15 to change anything. What announcement or API change prompted this question?

AppleEducate:
@stuartmorgan just watching all the keynotes from WWDC19. I'll see if I can find the link but with SwiftUI it really focused on context menus and with them coming to MacOS it would create the correct navigation menus and submenus. 

I don't think the API changed for `NSMenu` and `MSMenuItem`. But I do know they added context menus: https://developer.apple.com/documentation/swiftui/navigationview/3285966-contextmenu

stuartmorgan:
Since there are no plans to use SwiftUI in the macOS Flutter embedding, how it wraps portions of AppKit will have no impact on plugins in this repository.

GroovinChip:
Just want to chime in to say that on Windows if you press the 'escape' key while focusing on a TextField a strange character appears in the TextField. Screenshot:
[...]

PR #422, opened 2019-06-05
Add initial shortcut support to menubar for macOS

stuartmorgan:
Adds the ability to set shortcuts for menu items, using the same format
on the Dart side that will be used for Action shorcuts. Currently only
the macOS platform implementation uses them.

There are some limitations around non-ascii keys, which will be resolved
via a full generated keymap once this is folded into Flutter, but common
cases are special-cased, and more special-casing can be added as needed
by clients.

Adds various shortcuts to the testbed app, exercising various modifiers
and special cases.

Partially addresses #95

### DIFF
--- a/plugins/menubar/lib/src/menu_channel.dart
+++ b/plugins/menubar/lib/src/menu_channel.dart
@@ -14,6 +14,7 @@
 import 'dart:async';
 
 import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
 
 import 'menu_item.dart';
 
@@ -26,6 +27,7 @@ const String _kMenuChannelName = 'flutter/menubar';
 /// The argument to this method will be an array of map representations
 /// of menus that should be set as top-level menu items.
 const String _kMenuSetMethod = 'Menubar.SetMenu';
+
 /// The method name for the Dart-side callback called when a menu item is
 /// selected.
 //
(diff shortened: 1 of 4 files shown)


plugins/menubar/macos/Classes/MenubarPlugin.m
```diff
@@ -20,10 +20,82 @@
...
+      character = NSDeleteCharacter;
+      break;
+    default:
+      return nil;
+  }
+  return [NSString stringWithCharacters:&character length:1];
+}
+
+// Returns the NSEventModifierFlags of |modifiers|, a value from kShortcutKeyModifiers.
+static NSEventModifierFlags KeyEquivalentModifierMaskForModifiers(NSNumber *modifiers) {
```
krisgiesing: Do we need to handle the other modifiers here? (NSEventModifierFlagCapsLock, NSEventModifierFlagNumericPad, NSEventModifierFlagHelp, NSEventModifierFlagFunction)

plugins/menubar/lib/src/menu_channel.dart
```diff
@@ -154,6 +208,49 @@ class MenuChannel {
...
+    for (final key in shortcut.keys) {
+      if (key == LogicalKeyboardKey.meta) {
+        modifiers |= _shortcutModifierMeta;
+      } else if (key == LogicalKeyboardKey.shift) {
+        modifiers |= _shortcutModifierShift;
+      } else if (key == LogicalKeyboardKey.alt) {
+        modifiers |= _shortcutModifierAlt;
+      } else if (key == LogicalKeyboardKey.control) {
+        modifiers |= _shortcutModifierControl;
+      } else {
```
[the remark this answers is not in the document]
stuartmorgan: I'd like to have a real-world use case first; I can't think of having ever seen that as a menu item modifier.

plugins/menubar/macos/Classes/MenubarPlugin.m
stuartmorgan: This code only needs to handle the values that the platform channel can send, which is these four. If you mean do we need to add those on the Dart side too, same comment about actual use case applies to those too.

krisgiesing: My brief perusal of the docs suggested that numpad keys would be handled by a combination of number string and the NSEventModifierFlagNumericPad modifier, as opposed to the numpad keys having separate entries they way they are in LogicalKeyboardKey on the Dart side. It looks like this code can't distinguish between numpad number keys and regular numeric keys. Then there are similar questions about the other things that macOS considers modifiers which aren't in the set of allowed modifiers that this plugin defines.

But anyway, yeah, it's not clear how much people need that.

PR #423, opened 2019-06-06
Convert macOS example_plugin to Swift

### LINKED ISSUE #421: Convert macOS example_plugin to Swift
There should be at least one Swift plugin in the mix for testing. Since the Runner example for macOS uses Swift, it makes sense for the plugin example to be Swift as well.

stuartmorgan:
Uses the same structure as an iOS Flutter plugin using the Swift
template, with an ObjC interface wrapper.

This required adding use_frameworks to the Podfile, which will be added
to the template upstream as well.

Also adds the disable_input_output_paths that was added upstream to
address strict build issues with duplicate outputs.

Fixes #421

### DIFF
--- a/plugins/example_plugin/macos/Classes/ExamplePlugin.m
+++ b/plugins/example_plugin/macos/Classes/ExamplePlugin.m
@@ -14,22 +14,12 @@
 
 #import "ExamplePlugin.h"
 
+#import <example_plugin/example_plugin-Swift.h>
+
 @implementation FDEExamplePlugin
 
 + (void)registerWithRegistrar:(id<FlutterPluginRegistrar>)registrar {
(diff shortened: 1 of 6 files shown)


stuartmorgan:
Now that I've played with this more, I'm going to look at revisiting the plugin tooling structure for macOS. Given that the plan is to support only the Swift runner on macOS, I'd like to explore inverting the plugin structure for macOS so that plugins are Swift-first, with ObjC plugins being wrapped by a Swift API, rather than the reverse as we have now. I think that will give a more forward-compatible structure (assuming there aren't complications getting the CocoaPods flow to work correctly for that setup).

But this is useful as an interim step even if that change does happen soon.

Issue #434, opened 2019-06-17
[flutter/plugins] Support shared_preferences


stuartmorgan:
This is simple for macOS since it can use NSUserDefaults as on iOS. Investigation is needed to figure out what the best equivalents would be for Linux and Windows.

axellebot:
So you forked the flutter/plugins repository into a subfolder and they will be merged later, right ?

stuartmorgan:
There's no forking involved, the "plugins" in the `flutter_plugins` directory here contain only the desktop platform implementations. The are meant to be [combined with the actual plugins](https://github.com/google/flutter-desktop-embedding/tree/master/plugins/flutter_plugins).

stellarbear:
This is a must have desktop feature

axellebot:
> For Windows though, since the current plugin API is expected to completely change (right now it shares the Linux C++ API; I expect it to switch to C# later), that's probably premature; discussions about how to proceed with that are still in progress. In the short term, I think adding a Windows implementation of shared_preferences here, with the expectation that it would move to flutter/plugins later, would be reasonable.

@stuartmorgan ATM if we want the plugin to work on windows we need to add linux support with GTK since windows plugin are done through Linux  C++ lib, right ? Or I totally miss understood ?

stuartmorgan:
Sorry, I'm not following the question. Windows and Linux implementations are entirely unrelated, so no, a Linux implementation is not required for a Windows implementation to exist. Even on Linux I don't see where GTK would come into it since `shared_preferences` has no UI.

I recommend looking at the Windows example plugin to see the basics of implementing a Windows plugin with the C++ API.

axellebot:
I probably miss understood the purpose of GTK, I thought GTK was used as a platform SDK as we do with Android SDK and iOS SDK.

Glad to hear that you created #408 for windows example, I will take a look ! Thanks 👌

lucacasonato:
Do Windows and Linux have equivalents to NSUserDefaults? I presume we will have to write a json or xml file to disk (sorta like android shared preferences works under the hood). For Linux this can be in $XDG_CONFIG_HOME/bundle_id/shared_prefrences.json (https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). For Windows we can probably use %AppData%bundle_id/shared_prefrences.json. 

I haven't really done C++ stuff extensively so it is probably best that someone who knows more about this implements this.

stuartmorgan:
[...]

PR #435, opened 2019-06-17
Add shared_preferences implementation for macOS

stuartmorgan:
Implements shared_preferences_fde for macOS, for use with the
flutter/plugins shared_preferences plugin.

This is the macOS portion of #434

### DIFF
--- a/plugins/flutter_plugins/shared_preferences_fde/macos/Classes/SharedPreferencesPlugin.swift
+++ b/plugins/flutter_plugins/shared_preferences_fde/macos/Classes/SharedPreferencesPlugin.swift
@@ -1,0 +1,62 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import FlutterMacOS
+import Foundation
+
+public class SharedPreferencesPlugin: NSObject, FlutterPlugin {
+  public static func register(with registrar: FlutterPluginRegistrar) {
+    let channel = FlutterMethodChannel(
+      name: "plugins.flutter.io/shared_preferences",
+      binaryMessenger: registrar.messenger)
+    let instance = SharedPreferencesPlugin()
+    registrar.addMethodCallDelegate(instance, channel: channel)
+  }
+
+  public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+    let method = call.method
+    if method == "getAll" {
+      result(getAllPrefs())
+    } else if method == "setBool" ||
+              method == "setInt" ||
+              method == "setInt" ||
+              method == "setDouble" ||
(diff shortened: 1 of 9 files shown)


plugins/flutter_plugins/shared_preferences_fde/macos/Classes/SharedPreferencesPlugin.swift
```diff
@@ -0,0 +1,62 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import FlutterMacOS
+import Foundation
+
+public class SharedPreferencesPlugin: NSObject, FlutterPlugin {
+  public static func register(with registrar: FlutterPluginRegistrar) {
+    let channel = FlutterMethodChannel(
+      name: "plugins.flutter.io/shared_preferences",
```
franciscojma86: Should these strings (and below) be extracted to constants?

plugins/flutter_plugins/shared_preferences_fde/macos/Classes/SharedPreferencesPlugin.swift
```diff
@@ -0,0 +1,62 @@
...
+    let method = call.method
+    if method == "getAll" {
+      result(getAllPrefs())
+    } else if method == "setBool" ||
+              method == "setInt" ||
+              method == "setInt" ||
+              method == "setDouble" ||
+              method == "setString" ||
+              method == "setStringList" {
+      let arguments = call.arguments as! [String : Any]
```
franciscojma86: Should these be extracted with "if let"s? We've found that if a plugin crashes, the whole FDE crashes as well.

plugins/flutter_plugins/shared_preferences_fde/macos/Classes/SharedPreferencesPlugin.swift
stuartmorgan: I followed the flutter/plugins style for these, which at least in the plugins I was looking at tended to inline them. I think for simple plugins like this it doesn't matter much one way or the other, which is why I opted for that style rather than what we've been doing with the FDE plugins. When this moves to an official location I think we should have a written style on this.

plugins/flutter_plugins/shared_preferences_fde/macos/Classes/SharedPreferencesPlugin.swift
stuartmorgan: These arguments are constructed internally to the Dart plugin, not user input, so failing this check would be a serious bug in the plugin itself. We should decide how to handle that generally as well.

franciscojma86: Sounds good, thanks!

Issue #440, opened 2019-06-25
[MacOS/iOS plugins] MacOS plugins dependency breaks iOS cocoapod install


rxlabz:
I try to use the same flutter project for both macOS and iOS.
All my platform specific code is separated in dedicated packages ( desktop and mobile ).
The desktop package uses the filechooser plugin, and has it declared in his pubspec. 
When I add the desktop_package as dependencies in the project pubspec, I can't compile the iOS app.
The filechooser plugin is added in the `.flutter-plugins` file and when cocoa pod run, I got this error  : 

```
12:46:48.587 27 info flutter.tools     Fetching external sources
12:46:48.587 28 info flutter.tools     -> Fetching podspec for `Flutter` from `.symlinks/flutter/ios`
12:46:48.587 29 info flutter.tools     -> Fetching podspec for `file_chooser` from `.symlinks/plugins/file_chooser/ios`
12:46:48.587 30 info flutter.tools     [!] No podspec found for `file_chooser` in `.symlinks/plugins/file_chooser/ios`
```

The iOS app is compiled when I comment the desktop_package depency.

OS: iOS / MacOS
```
#flutter doctor
[✓] Flutter (Channel dev, v1.7.9, on Mac OS X 10.14.5 18F132, locale fr-FR)

[✓] Android toolchain - develop for Android devices (Android SDK version 28.0.3)
[✓] Xcode - develop for iOS and macOS (Xcode 10.2.1)
[✓] iOS tools - develop for iOS devices
[✓] Chrome - develop for the web
[✓] Android Studio (version 3.4)
[✓] Connected device (3 available)
```

Complete cocoa pod log error :

```
12:46:43.869 1 info flutter.tools Launching lib/main.dart on iPhone 7 Plus in debug mode...
12:46:47.953 2 info flutter.tools Running pod install...
12:46:48.587 3 info flutter.tools CocoaPods' output:
12:46:48.587 4 info flutter.tools ↳
12:46:48.587 5 info flutter.tools     pod /Users/rxlabz/dev/tools/flutter
12:46:48.587 6 info flutter.tools     pod /Users/rxlabz/dev/labz/flutter/flutter-xplatform/xp_app
12:46:48.587 7 info flutter.tools     pod /Users/rxlabz/dev/labz/flutter/flutter-xplatform/xp_app/lib/main.dart
[...]

PR #442, opened 2019-06-27
add a temporary fake ios plugin to each plugin. 

rxlabz:
Just some minimal "implementation"of iOS plugins as suggested here #440
This allows to compile iOS flutter apps even if some of their dependencies have themselves some desktop plugins dependencies.


plugins/color_panel/ios/Classes/ColorPanelPlugin.m
```diff
@@ -0,0 +1,20 @@
+#import "ColorPanelPlugin.h"
+
+@implementation ColorPanelPlugin
++ (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar>*)registrar {
+  FlutterMethodChannel* channel = [FlutterMethodChannel
```
stuartmorgan: There's no need for any of the code in this method.

plugins/color_panel/ios/Classes/ColorPanelPlugin.m
```diff
@@ -0,0 +1,20 @@
...
+      methodChannelWithName:@"color_panel"
+            binaryMessenger:[registrar messenger]];
+  ColorPanelPlugin* instance = [[ColorPanelPlugin alloc] init];
+  [registrar addMethodCallDelegate:instance channel:channel];
+}
+
+- (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
```
stuartmorgan: This method can be removed entirely.

plugins/color_panel/ios/color_panel.podspec
```diff
@@ -0,0 +1,21 @@
+#
+# To learn more about a Podspec see http://guides.cocoapods.org/syntax/podspec.html
+#
+Pod::Spec.new do |s|
+  s.name             = 'color_panel'
+  s.version          = '0.0.1'
+  s.summary          = 'temp fake color_panel plugin'
```
stuartmorgan: Please provide a clearer description (here and for `description`) so that someone looking at this podspec would understand what this is. Something like "No-op implementation of color_panel to avoid build issues on iOS"

plugins/color_panel/ios/color_panel.podspec
```diff
@@ -0,0 +1,21 @@
...
+  s.description      = <<-DESC
+temp fake color_panel plugin
+                       DESC
+  s.homepage         = 'http://example.com'
```
stuartmorgan: This and `author` should have real values, as with macOS.

plugins/color_panel/ios/Classes/ColorPanelPlugin.m
rxlabz: thanks for your feedback and suggestions. Updates are done.

rxlabz:
Done !
What do you think about adding the same kind of iOS placeholder in the example_plugin folder ?

stuartmorgan:
> What do you think about adding the same kind of iOS placeholder in the example_plugin folder ?

Someone using example_plugin may be developing a plugin that works on any platform, in which case they would want the skeletons from a standard Flutter plugin template, not a no-op implementation. And if they do, do they want Swift, or ObjC? (And for Android, Kotlin or Java?)

If other people want to experiment with making desktop-only plugins between now and when `flutter` is updated to handle plugins without mobile implementations, they can decide if/how they want to handle that for now. I would much rather they have to make that decision and understand what they are doing than have people duplicating a temporary hack without even knowing it.

hacker1024:
I still can't build for iOS:

```
    -> Fetching podspec for `window_size` from `.symlinks/plugins/window_size/ios`
    [!] No podspec found for `window_size` in `.symlinks/plugins/window_size/ios`
```

stuartmorgan:
@hacker1024 You need to update your Podfile; you should have received a migration warning from Flutter at some point warning you about that. See https://github.com/flutter/flutter/issues/45197#issuecomment-670602842

A project that's using an updated Podfile doesn't need iOS stubs, so I removed them all from this repository.

hacker1024:
Sorry - I missed the message due to it being above all the build output. Thanks.

PR #454, opened 2019-07-10
Remove launch call on FLEViewController

stuartmorgan:
Updates macOS Runners for
flutter/engine#9750

Projects are now configured via init, and there is no explicit launch
command on a view controller.


stuartmorgan:
Note: this is on top of 453 so you should only look at the last commit, not the whole diff. I'll land 453 and then rebase this before landing it, but I wanted to post both sides of this change at the same time, and 453 can't land until the engine rolls to pick up the other side of that change.

PR #458, opened 2019-07-14
implemented url_launcher_fde for Windows

csells:


csells:
I hate to use the old "it works on my machine," but it does! I do a "git clean -fx" locally, "flutter run" from testbed and it builds and runs just fine...

plugins/flutter_plugins/url_launcher_fde/windows/url_launcher_fde.cpp
```diff
@@ -0,0 +1,120 @@
...
+                                 // 'HINSTANCE' to 'int'
+    int status =
+        (int)ShellExecute(NULL, TEXT("open"), wurl.c_str(), NULL, NULL, SW_SHOWNORMAL);
+#pragma warning(pop)
+
+    if (status <= 32) {
+      std::ostringstream error_message;
+      error_message << "Failed to open " << url << ": ShellExecute error code "
+                    << status;
+      result->Error("open_error", error_message.str());
```
stuartmorgan: There needs to be a return after this, so |result| isn't called twice. It looks like I messed that up in the Linux plugin, which I'll fix right now.

testbed/windows/Runner.sln
```diff
@@ -10,6 +10,8 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Runner", "Runner.vcxproj",
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ExamplePlugin", "..\..\plugins\example_plugin\windows\ExamplePlugin.vcxproj", "{95E21B2C-C18A-4CED-8509-585CB2570FDE}"
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "url_launcher_fde", "..\..\plugins\flutter_plugins\url_launcher_fde\windows\url_launcher_fde.vcxproj", "{18F565CD-7BD5-459C-9FEB-2E2379E88105}"
```
stuartmorgan: Is having a duplicate UUID valid in VS? It might be a good idea to hand-edit the duplicated project to change the UUID.

plugins/flutter_plugins/url_launcher_fde/windows/.gitignore
```diff
@@ -0,0 +1,7 @@
+flutter/
+scripts/*.log
```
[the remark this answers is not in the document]
csells: my work produced flutter log files

plugins/flutter_plugins/url_launcher_fde/windows/url_launcher_fde.cpp
```diff
@@ -0,0 +1,120 @@
...
+      if (url_it != arguments.end()) {
+        url = url_it->second.StringValue();
+      }
+    }
+    if (url.empty()) {
+      result->Error("argument_error", "No URL provided");
+      return;
+    }
+
+    // Win32-specific code
```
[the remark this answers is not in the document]
csells: only meant to point out that in the sea of boilerplate code that this was the meat of the extension. I'll update the comment.

testbed/windows/Runner.sln
csells: good eye!

plugins/flutter_plugins/url_launcher_fde/windows/.gitignore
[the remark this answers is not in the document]
stuartmorgan: Do you mean crash log files? Those don't seem like something that needs to be in the .gitignore (and certainly not specifically to this folder).

testbed/windows/Runner.sln
stuartmorgan: After looking more I see that the one on the left isn't the ID of the referenced project, and doesn't seem to be unique already (the line just outside of context uses the same one for the app project too). So it seems like this is not an issue.

plugins/flutter_plugins/url_launcher_fde/windows/url_launcher_fde.vcxproj
```diff
@@ -0,0 +1,154 @@
...
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+    <Import Project="$(SolutionDir)flutter\Generated.props" />
+  </ImportGroup>
+  <PropertyGroup Label="UserMacros" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <LinkIncremental>true</LinkIncremental>
+    <IncludePath>$(VC_IncludePath);$(WindowsSDK_IncludePath);$(ProjectDir)flutter\cpp_client_wrapper\include;$(ProjectDir)flutter</IncludePath>
+    <LibraryPath>$(VC_LibraryPath_x64);$(WindowsSDK_LibraryPath_x64);$(NETFXKitsDir)Lib\um\x64;$(ProjectDir)flutter</LibraryPath>
+    <OutDir>$(SolutionDir)..\..\build\windows\$(Platform)\$(Configuration)\Plugins\</OutDir>
```
stuartmorgan: This and the next line are wrong; they have an extra `..\` relative to the example and to the release version below.

.gitignore
```diff
@@ -1 +1,3 @@
 .DS_Store
+.vscode/
+build/
```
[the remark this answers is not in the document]
stuartmorgan: My comment about the output directory for Debug in the project answers this question, I believe. Your build output shouldn't have been going here.

plugins/flutter_plugins/url_launcher_fde/windows/.gitignore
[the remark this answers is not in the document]
csells: unless they're in the .gitignore, we'll check them in, which doesn't seem right.

csells:
I think I screwed up the Linux build somehow... can you take a look?

plugins/flutter_plugins/url_launcher_fde/windows/.gitignore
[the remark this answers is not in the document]
stuartmorgan: Let's back up a bit: are you getting Flutter tool crash logs (e.g., `flutter_01.log`) in this folder every time you build? If so, that's a serious problem, and we shouldn't be masking it with .gitignore.

If not, and it was something that you happened to do locally at some point in your work making the patch, then the solution is just to not check in the file, as with any other random files you might happen to put in your workspace that aren't part of the patch.

stuartmorgan:
> I think I screwed up the Linux build somehow... can you take a look?

The file diffs show that you removed the executable bit from several of the CI scripts. Your patch shouldn't be changing those files.

csells:
Agreed. I'm not sure how to back out that change.

>

stuartmorgan:
Scripts reverted.

plugins/flutter_plugins/url_launcher_fde/windows/url_launcher_fde.vcxproj
csells: I believe this is fixed.

csells:
I think I applied all of your requested changed.

plugins/flutter_plugins/url_launcher_fde/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -0,0 +1,11 @@
+//
```
stuartmorgan: These `macos/` files shouldn't exist (and definitely shouldn't be added in a Windows patch regardless). Filed https://github.com/flutter/flutter/issues/37039 for the fact that they are created, but they should be reverted from this patch.

plugins/flutter_plugins/url_launcher_fde/windows/url_launcher_fde.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    // launch a URL on Windows
+    size_t size = url.length() + 1;
+    std::wstring wurl;
+    wurl.reserve(size);
+    size_t outSize;
+    mbstowcs_s(&outSize, &wurl[0], size, url.c_str(), size - 1);
+
+#pragma warning(push)
+#pragma warning(disable : 4311)  // warning C4311: 'type cast': pointer
```
stuartmorgan: Are these warning-disables necessary now that you are `reinterpret_cast`ing?

csells: unfortunately yes...

PR #459, opened 2019-07-14
Updates to Xcode project to support flutter assemble

jonahwilliams:
Two changes to the example and testbed xcodeprojects to support the changes in the flutter tools build system.

- Refer to flutter_assets relative to the $FLUTTER_BUILD_DIR config value. The assets are now placed in a build directory that is scoped specifically for that build instead of shared + always regenerated.
- Provide an xcfilelist to the InputFileList and ouputFileList properties of the bundle script. This will allow xCode to completely skip the script phase when the inputs/outputs have not changed. The xcfilelists are generated by the fluter_tool. This allows repeated builds without changes to be almost instant.

For more context see: https://github.com/flutter/flutter/pull/36164 , the corresponding flutter tools change.

EDIT: added link to corresponding flutter tools change.


example/macos/.gitignore
```diff
@@ -4,3 +4,7 @@ xcuserdata
 Flutter/ephemeral/
 # Created by CocoaPods for plugins.
 Pods/
+
+# Create by flutter tool for XCode
+FlutterInputs.xcfilelist
```
jonahwilliams: I didn't really know where to put these files so I stuck them in the root.

example/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -220,12 +220,8 @@
 			buildActionMask = 2147483647;
 			files = (
 			);
-			inputFileListPaths = (
```
jonahwilliams: Unrelated change, I can revert

testbed/macos/Podfile.lock
```diff
@@ -55,4 +55,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 8e22ef6b498b5efcd617873b5b5fe67bc84c66ed
 
-COCOAPODS: 1.7.1
```
jonahwilliams: Similar unrelated change

example/macos/.gitignore
stuartmorgan: If they are created by `flutter`, and not intended to be checked in, they should be in Flutter/ephemeral.

testbed/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -26,7 +26,7 @@
 		33CC10F32044A3C60003C045 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 33CC10F22044A3C60003C045 /* Assets.xcassets */; };
 		33CC10F62044A3C60003C045 /* MainMenu.xib in Resources */ = {isa = PBXBuildFile; fileRef = 33CC10F42044A3C60003C045 /* MainMenu.xib */; };
 		33CC11132044BFA00003C045 /* FlutterWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33CC11122044BFA00003C045 /* FlutterWindow.swift */; };
-		33CC112F204626C80003C045 /* flutter_assets in Resources */ = {isa = PBXBuildFile; fileRef = 33CC112C20461AD40003C045 /* flutter_assets */; };
+		33CC112F204626C80003C045 /* flutter_assets in Resources */ = {isa = PBXBuildFile; fileRef = 33CC112C20461AD40003C045 /* $flutter_assets */; };
```
stuartmorgan: Is this cruft from earlier experimentation? I wouldn't expect the name to have changed.

jonahwilliams: Fixed

testbed/macos/Podfile.lock
jonahwilliams: Revert

example/macos/Runner.xcodeproj/project.pbxproj
jonahwilliams: Reverted

example/macos/.gitignore
jonahwilliams: File paths moved, still need to update tool pr

jonahwilliams:
assemble for macOS should be landing shortly in the framework side, I'll re-trigger the build here shortly.

stuartmorgan:
@jonahwilliams  I landed this since things were completely broken without it, but CI for that last commit shows both `--release` build of `example` and the (debug) build of `testbed` are broken still. I likely won't have time to investigate today; could you take a look? If it's going to be non-trivial to fix, we should revert the assemble change and this change until the failures can be addressed.

jonahwilliams:
I have a fix here: https://github.com/flutter/flutter/pull/37365. That's my bad, I checked flutter run but not build --release

PR #460, opened 2019-07-15
Fix double-result on error in Linux url_launcher

stuartmorgan:
Adds a missing return in error handling, leading to a path that calls
|result| twice.

Also fixes a comment that was left over from an earlier plugin that kept
a channel reference.


krisgiesing:
Nit, the commit message has a typo (doule -> double).

PR #486, opened 2019-08-15
Use 10.11 as deployment target

stuartmorgan:
See https://github.com/flutter/flutter/issues/33200


stuartmorgan:
I was going to land this after the buildroot->engine->flutter cascade is done for the engine part of this change, which will be a while. So you first.

PR #491, opened 2019-08-20
Update plugins for platform yaml changes

stuartmorgan:
Updates all plugins to conform to the new plugin schema in
https://github.com/flutter/flutter/pull/38632


stuartmorgan:
Added a rev of the min compatible Flutter version (which I have, predictably, forgotten to do almost all of the time for these kinds of breaking changes).

PR #492, opened 2019-08-20
Remove xcfilelist input and outputs

jonahwilliams:
as discussed in https://github.com/flutter/flutter/issues/38863, these won't work with multiple configurations - though the flutter_tool driven caching will still apply so the loss isn't huge.


stuartmorgan:
LGTM as a quick fix, but I like jmagman's suggestion of different paths as a longer term solution.

jonahwilliams:
agreed

PR #493, opened 2019-08-21
[Windows] Update example and testbed to use the native win32 Windows engine

clarkezone:
This change is WIP for the consumption of the following engine PR:

https://github.com/flutter/engine/pull/11327

This PR shouldn't be accepted until native win32 is the default Windows embedder: https://github.com/flutter/flutter/pull/39836


common/windows/win32_window.cc
```diff
@@ -0,0 +1,150 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "win32_window.h"
+#include "shellscalingapi.h"
+
+// the Windows DPI system is based on this
```
stuartmorgan: Anonymous namespace for this and the helper to give them internal linkage.

common/windows/win32_window.cc
```diff
@@ -0,0 +1,150 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "win32_window.h"
```
stuartmorgan: Newline after this, per style guide.

common/windows/win32_window.cc
```diff
@@ -0,0 +1,150 @@
...
+#include "shellscalingapi.h"
+
+// the Windows DPI system is based on this
+// constant for machines running at 100% scaling.
+constexpr int base_dpi = 96;
+
+// Scale helper to convert logical scaler values to physical using passed in
+// scale factor
+int Scale(int source, double scale_factor) {
+  return static_cast<int>(static_cast<double>(source * scale_factor));
```
stuartmorgan: The inner expression is already a double, so no need to cast it to a double.

common/windows/win32_window.cc
```diff
@@ -0,0 +1,150 @@
...
+bool Win32Window::CreateAndShow(const char *title, const unsigned int x,
+                                const unsigned int y, const unsigned int width,
+                                const unsigned int height) {
+  Destroy();
+
+  WNDCLASS window_class = ResgisterWindowClass(title);
+
+  // New windows are created on the default monitor.  Window sizes are specified
+  // to the OS in physical pixels, hence to ensure a consistent size to will
+  // treat the width height passed in to this function as logical pixels and
```
stuartmorgan: This seems like something that should be in the header comment, not here.

common/windows/win32_window.cc
```diff
@@ -0,0 +1,150 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "win32_window.h"
+#include "shellscalingapi.h"
+
+// the Windows DPI system is based on this
+// constant for machines running at 100% scaling.
+constexpr int base_dpi = 96;
```
stuartmorgan: kBaseDpi

common/windows/win32_window.cc
```diff
@@ -0,0 +1,150 @@
...
+
+  // New windows are created on the default monitor.  Window sizes are specified
+  // to the OS in physical pixels, hence to ensure a consistent size to will
+  // treat the width height passed in to this function as logical pixels and
+  // scale to appropriate for the default monitor.
+  HMONITOR defaut_mon = MonitorFromWindow(nullptr, MONITOR_DEFAULTTOPRIMARY);
+  UINT dpi_x = 0, dpi_y = 0;
+  GetDpiForMonitor(defaut_mon, MDT_EFFECTIVE_DPI, &dpi_x, &dpi_y);
+
+  double scaleFactor = static_cast<double>(dpi_x) / base_dpi;
```
stuartmorgan: scale_factor

common/windows/win32_window.cc
```diff
@@ -0,0 +1,150 @@
...
+
+  auto window = CreateWindow(
```
stuartmorgan: I don't know offhand what the type of this is, so I would prefer not using `auto`

common/windows/win32_window.cc
```diff
@@ -0,0 +1,150 @@
...
+      window_class.lpszClassName, title, WS_OVERLAPPEDWINDOW | WS_VISIBLE,
+      Scale(x, scaleFactor), Scale(y, scaleFactor), Scale(width, scaleFactor),
+      Scale(height, scaleFactor), nullptr, nullptr, window_class.hInstance,
+      this);
+  if (window == nullptr) {
```
stuartmorgan: `return window != nullptr`?

common/windows/win32_window.cc
```diff
@@ -0,0 +1,150 @@
...
+                            LPARAM const lparam) noexcept {
+  int xPos = 0, yPos = 0;
+  UINT width = 0, height = 0;
+  auto window =
+      reinterpret_cast<Win32Window *>(GetWindowLongPtr(hwnd, GWLP_USERDATA));
+
+  if (window != nullptr) {
+    switch (message) {
+      case WM_DESTROY:
+        /*window->OnClose();*/
```
stuartmorgan: ?

common/windows/win32_window.cc
```diff
@@ -0,0 +1,150 @@
...
+
+LRESULT
+Win32Window::MessageHandler(HWND hwnd, UINT const message, WPARAM const wparam,
+                            LPARAM const lparam) noexcept {
+  int xPos = 0, yPos = 0;
+  UINT width = 0, height = 0;
+  auto window =
+      reinterpret_cast<Win32Window *>(GetWindowLongPtr(hwnd, GWLP_USERDATA));
+
+  if (window != nullptr) {
```
stuartmorgan: I'd prefer early return on `nullptr` rather than deeper nesting on the bulk of the logic.

common/windows/win32_window.h
```diff
@@ -0,0 +1,92 @@
...
+#ifndef WIN32_WINDOW_H_
+#define WIN32_WINDOW_H_
+
+#include <Windows.h>
+#include <Windowsx.h>
+
+#include <memory>
+#include <string>
+
+// A class abstraction for a high DPI aware Win32 Window.  Intended to be
```
stuartmorgan: Nit: "high DPI-aware"

common/windows/win32_window.h
```diff
@@ -0,0 +1,92 @@
...
+// inherited from by classes that wish to specialize with custom
+// rendering and input handling
+class Win32Window {
+ public:
+  Win32Window();
+  ~Win32Window();
+
+  // Creates and shows a win32 window with |title| and position and size using |x|,
+  // |y|, |width| and |height|
```
stuartmorgan: Document return value.

common/windows/win32_window.h
```diff
@@ -0,0 +1,92 @@
...
+  bool CreateAndShow(const char *title, const unsigned int x, const unsigned int y,
+                  const unsigned int width, const unsigned int height);
```
stuartmorgan: I'm slightly confused about the intended use with this being separate from the constructor. E.g., is calling this multiple times valid? If so, what does it do?

common/windows/win32_window.h
```diff
@@ -0,0 +1,92 @@
...
+
+  // Release OS resources asociated with window.
```
stuartmorgan: Can you elaborate on what makes this conceptually different from the destructor?

common/windows/win32_window.h
```diff
@@ -0,0 +1,92 @@
...
+  void Destroy();
+
+  void SetChildContent(HWND content);
```
stuartmorgan: Needs a comment.

common/windows/win32_window.h
```diff
@@ -0,0 +1,92 @@
...
+
+ protected:
+
```
stuartmorgan: Remove blank line for consistency.

common/windows/win32_window.h
```diff
@@ -0,0 +1,92 @@
...
+  // OS callback called by message pump.  Handles the WM_NCCREATE message which
+  // is passed when the non-client area is being created and enables automatic
+  // non-client DPI scaling so that the non-client area automatically
+  // responsponds to changes in DPI.  All other messages are handled by
+  // MessageHandler.
+  static LRESULT CALLBACK WndProc(HWND const window, UINT const message,
+                                  WPARAM const wparam,
+                                  LPARAM const lparam) noexcept;
+
+  // TODO
```
stuartmorgan: ^^^

common/windows/win32_window.h
```diff
@@ -0,0 +1,92 @@
...
+  static BOOL CALLBACK EnumChildProc(HWND child_window, LPARAM lParam);
+
+  // Processes and route salient window messages for mouse handling,
```
stuartmorgan: Nit: routes

common/windows/win32_window.h
```diff
@@ -0,0 +1,92 @@
...
+
+  // When WM_DPICHANGE resizes the window to the new suggested
+  // size and notifies inheriting class.
+  LRESULT
+  HandleDpiChange(HWND hWnd, WPARAM wParam, LPARAM lParam);
+
+  //// Called when the DPI changes either when a
+  //// user drags the window between monitors of differing DPI or when the user
+  //// manually changes the scale factor.
+  // virtual void OnDpiScale(UINT dpi) = 0;
```
stuartmorgan: ?

common/windows/win32_window.h
```diff
@@ -0,0 +1,92 @@
...
+
+  // window handle for hosted content.
+  HWND child_content_ = nullptr;
+
+  // Member variable to hold the window title.
+  const char* window_class_name_;
+
+  //// Member variable referencing an instance of dpi_helper used to abstract
+  /// some / aspects of win32 High DPI handling across different OS versions.
+  // std::unique_ptr<Win32DpiHelper> dpi_helper_ =
```
stuartmorgan: ?

common/windows/win32_window.h
```diff
@@ -0,0 +1,92 @@
...
+#include <Windows.h>
+#include <Windowsx.h>
+
+#include <memory>
+#include <string>
+
+// A class abstraction for a high DPI aware Win32 Window.  Intended to be
+// inherited from by classes that wish to specialize with custom
+// rendering and input handling
+class Win32Window {
```
stuartmorgan: How would setting title, icon, size, etc. work here, given that the handle is internal?

example/windows/win32_window.cc
```diff
@@ -0,0 +1,240 @@
...
+  // Assume Windows 10 1703 or greater for DPI handling.  When running on a
+  // older release of Windows where this context doesn't exist, DPI calls will
+  // fail and Flutter rendering will be impacted until this is fixed.
+  // To handle downlevel correctly, dpi_helper must use the most recent DPI
+  // context available should be used: Windows 1703: Per-Monitor V2, 8.1:
+  // Per-Monitor V1, Windows 7: System See
+  // https://docs.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows
+  // for more information.
+
+  // TODO the calling applicaiton should participate in setting the DPI.
```
stuartmorgan: Needs updating for the fact that this is now the calling application.

common/windows/win32_window.h
clarkezone: Handle exposed in next commit

example/windows/flutter_embedder_example.cpp
```diff
@@ -16,7 +16,9 @@
 #include <string>
 #include <vector>
 
-#include "flutter/flutter_window_controller.h"
+#include "flutter/flutter_view_controller.h"
+
+#include "..\..\common\windows\win32_window.h"
```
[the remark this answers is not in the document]
clarkezone: I don't think the window code belongs in the framework, it's part of the sample IMO.  The fact that it's non-trivial is a trait of win32.  We could put it in the wrapper or copy into every sample.  I will defer addressing this until after we have the custom task runner done and are ready to switch over to the native windows implementation.

common/windows/win32_window.h
```diff
@@ -0,0 +1,95 @@
...
+ public:
+  Win32Window();
+  ~Win32Window();
+
+  // Creates and shows a win32 window with |title| and position and size using
+  // |x|, |y|, |width| and |height|
+  bool CreateAndShow(const char *title, const unsigned int x,
+                     const unsigned int y, const unsigned int width,
+                     const unsigned int height);
```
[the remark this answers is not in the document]
clarkezone: There is RECT which is top left right bottom.  A bit annoying as you can't express what you really want (width and height).  Can switch to that if you like but it will create even more boilerplate windowing code as we will need to reconstruct width.& height to correctly preserve DPI handling.

stuartmorgan:
Synced with master, fixed all the references to `_glfw` files. I'll do some more updates to address the remaining things (eliminating common, maybe adding some point/size classes for the APIs, etc.), but at this point applying this and https://github.com/flutter/flutter/pull/39836 locally should work.

example/windows/win32_window.h
```diff
@@ -0,0 +1,132 @@
...
+  Win32Window();
+  ~Win32Window();
+
+  // Initializes and shows window with |title| and position and size using |x|,
+  // |y|, |width| and |height|
+  void Initialize(const char* title,
+                  const unsigned int x,
+                  const unsigned int y,
+                  const unsigned int width,
+                  const unsigned int height);
```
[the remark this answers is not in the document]
stuartmorgan: For now I'm going to add a Point and Size to this class to make things a bit more self-documenting. Not an ideal solution, but we can easily iterate later.

PR #496, opened 2019-08-22
Update Linux apps for incremental GLFW runloop

stuartmorgan:
Switches to the new API from https://github.com/flutter/engine/pull/11368 and moves the GTK event handling into the application level of 'testbed'. 'example' doesn't do that handling since it doesn't use plugins. (Longer term a specific toolkit like GTK will be a core requirement of the library, so this divergence won't exist; for now this makes the simple example easier to build and run.)

Note: Since the run-forever version in the C++ wrapper is still present (to avoid this change being a hard compilation break), and Windows didn't need special handling beyond what that does, Windows is not updated; it will be changing to a new API soon anyway, and leaving it avoids collisions with that work.

### DIFF
--- a/example/linux/Makefile
+++ b/example/linux/Makefile
@@ -20,6 +20,11 @@
 BINARY_NAME=flutter_desktop_example
 # The C++ code for the embedder application.
 SOURCES=flutter_embedder_example.cc
+# Extra flags (e.g., for library dependencies).
+SYSTEM_LIBRARIES=
+EXTRA_CXXFLAGS=
+EXTRA_CPPFLAGS=$(patsubst -I%,-isystem%,$(shell pkg-config --cflags $(SYSTEM_LIBRARIES)))
+EXTRA_LDFLAGS=$(shell pkg-config --libs $(SYSTEM_LIBRARIES))
 
 # Default build type. For a release build, set BUILD=release.
 # Currently this only sets NDEBUG, which is used to control the flags passed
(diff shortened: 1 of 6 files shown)


example/linux/Makefile
```diff
@@ -20,6 +20,11 @@
 BINARY_NAME=flutter_desktop_example
 # The C++ code for the embedder application.
 SOURCES=flutter_embedder_example.cc
+# Extra flags (e.g., for library dependencies).
```
stuartmorgan: The changes to this makefile are just because I like to minimize differences between the structure of `example` and `testbed`'s builds for clarity when people diff.

testbed/lib/main.dart
```diff
@@ -38,6 +38,7 @@ void main() {
 
   // Try to resize and reposition the window to be half the width and height
   // of its screen, centered horizontally and shifted up from center.
+  WidgetsFlutterBinding.ensureInitialized();
```
stuartmorgan: This fixes an assertion with trying to use the window size plugin too early. Presumably unrelated, so I could break it out if you prefer, but since it was a trivial fix that I found while making this change I didn't want to stop, switch branches, retest, etc.

stuartmorgan: Verified on macOS and moved to https://github.com/google/flutter-desktop-embedding/pull/497. I'll remove from this PR before landing.

PR #498, opened 2019-08-22
Stop copying flutter_assets to the main bundle

stuartmorgan:
https://github.com/flutter/engine/pull/11380 adds support for looking up
flutter_assets from App.framework, which the 'flutter' tool packages, so
we no longer need to copy them into the main bundle.

Also does some minor project cleanup around the addition of
App.framework to make it consisent with FlutterMacOS (moved to Flutter
rather than Frameworks, and removes an accidental addition of a
reference to the on-disk Flutter folder in 'example').


stuartmorgan:
Needs to land after the engine change has landed and rolled though to the Flutter tree, obviously.

PR #511, opened 2019-08-27
bugfix: #31268

shoryukenn:
issuse: https://github.com/flutter/flutter/issues/31268

Specify the link options in both modes explicitly, adapting different codes for different options.

### DIFF
--- a/example/windows/flutter_embedder_example.cpp
+++ b/example/windows/flutter_embedder_example.cpp
@@ -24,6 +24,17 @@
 #include <windows.h>
 #undef CreateWindow
 
+#include <fcntl.h>
+#include <io.h>
+#include <shellapi.h>
+#include <userenv.h>
+#include <psapi.h>
+
+#include <ios>
+#include <limits>
+#include <iostream>
+#pragma warning(disable: 4996)
+
 namespace {
 
 // Returns the path of the directory containing this executable, or an empty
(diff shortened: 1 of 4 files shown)


shoryukenn:
/subsystem:windows/conosle is not tied to debug/release.

About console:
1. The AllocConsole/AttachConsole allows the current application to have a new command line window.
2. You can redirect stdout to any console (like freopen) via AttachConsole(ATTACH_PARENT_PROCESS or pid).

Usually, we use the second, which means that example.vcxproj has /subsystem:windows by default. In the main function, we use AttachConsole(ATTACH_PARENT_PROCESS) to redirect stdout to the parent process (such as visual studio debug output or flutter run --release).

shoryukenn:
I modified the logic, the code is from Chromium, the project property is changed to /subsystem:windows, and the console--force-console is used to control whether to create the console. At the same time, if the parent process has a console, it will be attached directly. I think this is more consistent, what do you think?

example/windows/flutter_embedder_example.cpp
```diff
@@ -44,9 +55,88 @@ std::string GetExecutableDirectory() {
...
+}
+
 }  // namespace
 
-int main(int argc, char **argv) {
+
+int APIENTRY wWinMain(HINSTANCE instance, HINSTANCE prev, wchar_t * lpCmdLine, int) {
+
+  // 1. if parent or current process has console, attach it.
+  // 2. if --force-console exist && parent or current process have not conole,
```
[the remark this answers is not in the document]
shoryukenn: ok, will done.  I think your opinion is correct. According to our discussion, we only need one or two lines of code.

shoryukenn:
@stuartmorgan ,There is a scenario we have to confirm: 
What do we expect when Visual Studio debug example.exe?

example/windows/flutter_embedder_example.cpp
```diff
@@ -46,7 +46,14 @@ std::string GetExecutableDirectory() {
 
 }  // namespace
 
-int main(int argc, char **argv) {
+int APIENTRY wWinMain(HINSTANCE instance, HINSTANCE prev, wchar_t *, int) {
+
+  if (AttachConsole(ATTACH_PARENT_PROCESS)) {
+    FILE *p_fCout = nullptr;
```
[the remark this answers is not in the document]
shoryukenn: done

shoryukenn:
> I tried this locally, and it completely broke `flutter run`. It's not attaching to the observatory, so things like `q` and `r` don't do anything. It's critical that the Flutter tooling continue to work.
> 
> > What do we expect when Visual Studio debug example.exe?
> 
> Can you elaborate on the question? I'm not sure what specific behavior you mean.

I think when I use WinDBG or VS(f5) to start FDE.exe, generating a Console will be more friendly to developers.  So do we need to add a logic ? such as:
If (::IsDebuggerPresent()) {
   AllocConsole();
}

stuartmorgan:
> I think when I use WinDBG or VS(f5) to start FDE.exe, generating a Console will be more friendly to developers.

I see; I wasn't at my Windows machine, and had forgotten that VS doesn't itself act as a console for the running application (as compared to Xcode on macOS). I agree that if we can easily and reliably detect that case, showing a console would be helpful.

shoryukenn:
> > I think when I use WinDBG or VS(f5) to start FDE.exe, generating a Console will be more friendly to developers.
> 
> I see; I wasn't at my Windows machine, and had forgotten that VS doesn't itself act as a console for the running application (as compared to Xcode on macOS). I agree that if we can easily and reliably detect that case, showing a console would be helpful.

DONE.

example/windows/flutter_embedder_example.cpp
```diff
@@ -46,7 +46,16 @@ std::string GetExecutableDirectory() {
 
 }  // namespace
 
-int main(int argc, char **argv) {
+int APIENTRY wWinMain(HINSTANCE instance,
+                      HINSTANCE prev,
+                      wchar_t *command_line,
+                      int show_command) {
+  // attach to console of 'flutter run'
```
stuartmorgan: Nit: s/attach/Attach/, and add missing period at the end. The style guide requires that comments be punctuated as sentences.

shoryukenn: > I don't currently have access to my Windows machine. Did you confirm that this version works correctly with `flutter run`, attaching to the observatory and accepting/processing key input to trigger things like hot reload?

yes, I'm sure. It works fine in both scenarios (flutter run or VS debugger).
WXWorkCapture_15671489902245
WXWorkCapture_1567149044653

shoryukenn: > Nit: s/attach/Attach/, and add missing period at the end. The style guide requires that comments be punctuated as sentences.

DONE.

PR #518, opened 2019-08-29
dummy android module for file_chooser plugin

hpoul:
in #440 you mention to be open for adding dummy modules for android&ios to enable mobile compilation.. this adds an empty android module for file_chooser, just to make gradle happy.

If you like I can add it also to the other plugins (it would only require customizing the package name in `AndroidManifest.xml` for each plugin i guess). Although i'm not sure if dummy plugins will be made obsolete by https://github.com/flutter/flutter/pull/38632 ?

### DIFF
--- a/plugins/file_chooser/android/build.gradle
+++ b/plugins/file_chooser/android/build.gradle
@@ -1,0 +1,6 @@
+apply plugin: 'com.android.library'
+
+android {
+    compileSdkVersion 28
+}
+
--- a/plugins/file_chooser/android/src/main/AndroidManifest.xml
+++ b/plugins/file_chooser/android/src/main/AndroidManifest.xml
@@ -1,0 +1,4 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+  package="fde.file_chooser">
+</manifest>
+


stuartmorgan:
> Although i'm not sure if dummy plugins will be made obsolete by flutter/flutter#38632 ?

That's part of the larger effort, but I don't think it changes anything related to gradle. It also shouldn't be necessary to fix this problem for Android though, since unlike iOS Android has a required key even before that PR; the plugin registrant generation for Android already skips any plugin without a `androidPackage`. It sounds like similar logic is needed for gradle-related paths, and that will be true with the new manifest as well.

I'm okay taking this if necessary, but first I'd like to see if there's a quick fix that could be made on the Flutter side that would resolve Android build issues immediately (and also in the new-manifest world). What's the error specifically?

hpoul:
The actual gradle error is:
`ERROR: Could not get unknown property 'android' for project ':file_chooser' of type org.gradle.api.Project.`

It's from https://github.com/flutter/flutter/blob/master/packages/flutter_tools/gradle/flutter.gradle#L653-L665 I'd assume there is also some way of patching flutter.gradle, but I don't fully grasp how the plugins are included, i've tried to patch flutter.gradle by skipping adding gradle plugin tasks, but had no luck with that, so i figured adding a mock gradle file to the plugin was the easier way for now.. especially since there is already a dummy iOS plugin :-)

stuartmorgan:
I'll take a look sometime in the next couple of days, and if I can't quickly figure out how to omit the gradle step for plugins that don't support Android we can proceed with the dummy plugin.

stuartmorgan:
It doesn't look like there's an obvious quick fix that won't require some discussion/planning about how to handle `.flutter-plugins` in general. Filed https://github.com/flutter/flutter/issues/39657 for that, and I'll merge this. Thanks!

PR #527, opened 2019-09-03
Only Display Supported Widget Functions on Flutter View

canewsin:
- Since SharedPrefs are not supported on windows platform clicking them throwing errors on console. so show them only on supported platforms.
Same for ColorPanel, FileChooser Plugin.
 - Retargetting vsproj solution files to latest version(v142) currently (v141) including supported plugin solution files
 - Updated Dart SDK  minimum Constraints to 2.2.2
since `if` in list requires mimimum 2.2.2 version of dart sdk and we also use master branch of flutter so this most of the time contains latest stable version of dart sdk

### DIFF
--- a/testbed/lib/main.dart
+++ b/testbed/lib/main.dart
@@ -24,6 +24,7 @@ import 'package:example_flutter/keyboard_test_page.dart';
 import 'package:example_plugin/example_plugin.dart' as example_plugin;
 import 'package:file_chooser/file_chooser.dart' as file_chooser;
 import 'package:menubar/menubar.dart';
+import 'package:path_provider/path_provider.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 import 'package:url_launcher/url_launcher.dart' as url_launcher;
 import 'package:window_size/window_size.dart' as window_size;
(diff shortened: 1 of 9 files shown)


testbed/lib/main.dart
```diff
@@ -226,13 +226,14 @@ class _MyHomePage extends StatelessWidget {
...
         title: Text(title),
         actions: <Widget>[
-          new IconButton(
-            icon: new Icon(Icons.color_lens),
-            tooltip: 'Change theme color',
-            onPressed: () {
-              _changePrimaryThemeColor(context);
-            },
-          ),
+          if (!Platform.isWindows)
```
stuartmorgan: Can you explain what problem you are trying to solve here? Per [the README](https://github.com/google/flutter-desktop-embedding/blob/master/testbed/README.md) `testbed` is intended for use by people working on FDE. Hiding the unimplemented buttons just adds a step for anyone working on adding Windows support, so I don't see the benefit.

canewsin: Ok instead of Hiding i will try to implement them when i got time

PR #528, opened 2019-09-03
Added Multiple Solution Targets for VS proj file

canewsin:
Currently for only existed implementations of windows side plugins

PS:
i don't change the characterset on properties file i see that example solution file uses multibyte https://github.com/google/flutter-desktop-embedding/blob/master/example/windows/Runner.vcxproj#L25
but example_plugin uses unicode characterset is this change forgotten ?

### DIFF
--- a/plugins/example_plugin/windows/ExamplePlugin.vcxproj
+++ b/plugins/example_plugin/windows/ExamplePlugin.vcxproj
@@ -15,7 +15,7 @@
     <ProjectGuid>{95E21B2C-C18A-4CED-8509-585CB2570FDE}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>ExamplePlugin</RootNamespace>
-    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


plugins/example_plugin/windows/ExamplePlugin.vcxproj
```diff
@@ -15,19 +15,21 @@
     <ProjectGuid>{95E21B2C-C18A-4CED-8509-585CB2570FDE}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>ExamplePlugin</RootNamespace>
-    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
```
stuartmorgan: This breaks the bots; is it VS 2019+ only?

Also, apparently my comment on the previous PR was ambiguous, I meant for this change and the toolset changes to each be their own separate PRs, not split out into one new combined PR. The SDK change is less straightforward (more so than I thought, apparently), while the PlatformToolset change is trivial and doesn't require discussion.

stuartmorgan:
> example solution file uses multibyte but example_plugin uses unicode characterset is this change forgotten ?

I didn't make the initial runner project, so I'm not sure why it's set to multibyte. The plugin project used whatever the default was when I made a new project. Based on very limited research it seems like unicode is the better option, so another PR to make all projects use unicode would be good.

PR #529, opened 2019-09-03
Switch characterset to Unicode in solution props of example

canewsin:


canewsin:
i thought same issue first but i didn't follow, i thought its better position to start modifing this repo after 
https://github.com/google/flutter-desktop-embedding/pull/493 this is merged so that we can finalise what characterset to use in projects

PR #530, opened 2019-09-03
Disable C++ exceptions in example plugin

stuartmorgan:
This reduces the number of changes necessary for enabling CRL in a
plugin, and having C++ exceptions enabled isn't necessarily desirable
anyway.

Applies the same change to url_launcher_fde since with some few plugins
at the moment, people may use that as a template instead.


stuartmorgan:
The ginormous diff in url_launcher_fde.vcxproj is because it was apparently checked in without Windows line endings, so when VS rewrote it every line changed. I verified with `git diff -w` that it's just the two added lines of real diff.

PR #532, opened 2019-09-03
Switch plugin declaration for path_provider_fde

stuartmorgan:
The PR adding this plugin was posted before the change to pubspec.yaml
but landed after, so it doesn't work with the current version of
Flutter.


stuartmorgan:
Landing this TBR since it's a trivial change and fixes live bustage.

PR #533, opened 2019-09-03
Make the macOS example more distribution-friendly

stuartmorgan:
Moves the example closer to what we would want to ship as a template:
- Switches the signing settings back to match a default project (undoing
  an earlier change that disabled signing since signing wasn't a supported
  configuration at the time).
- Adds entitlements that are different for debug/profile vs release that
  have minimal exceptions for debug mode that should cover both
  sandboxing (incoming network requests, for observatory) and hardening
  if enabled (JIT).
- Turn on sandboxing by default, so that applications start as secure as
  possible, and developers need to decide what extra permissions they
  need.

The last is slightly experimental; it may be that there are unforeseen
issues with running in sandboxed mode that will require longer term
solutions. Turning it on by default allows us to gather feedback.

Part of flutter/flutter#32416


stuartmorgan:
Added some initial docs since this is likely to be a new source of potential confusion.

PR #535, opened 2019-09-05
Add inputs, outputs, and tripwire

jonahwilliams:
Adds input and output files lists back to xcode project. Adds script phase that unconditionally touches a tripwire file to force flutter to check the build again.

Confirmed locally that this is enough to force rebuilding without undoing the flutter caching.

### DIFF
--- a/example/macos/Flutter/flutter_config.sh
+++ b/example/macos/Flutter/flutter_config.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+# This is a generated file; do not edit or check into version control.
+
+touch Flutter/ephemeral/tripwire
--- a/example/macos/Flutter/flutter_config.sh
+++ b/example/macos/Flutter/flutter_config.sh
@@ -1,4 +1,0 @@
-#!/bin/sh
-# This is a generated file; do not edit or check into version control.
-
-touch Flutter/ephemeral/tripwire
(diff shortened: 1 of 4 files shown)


example/macos/Flutter/flutter_config.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+# This is a generated file; do not edit or check into version control.
+
+touch Flutter/ephemeral/tripwire
```
jonahwilliams: This script should probably go somewhere else

stuartmorgan: Why is it a separate script and not part of the script step that calls macos_assemble? Or even better, internal to macos_assemble so we can change it later without any project changes?

jonahwilliams: Ahh duh, that is much easier :)

example/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -257,16 +257,19 @@
...
 			);
 			outputFileListPaths = (
+				"Flutter/ephemeral/FlutterOutputs.xcfilelist"
 			);
 			outputPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "\"$FLUTTER_ROOT\"/packages/flutter_tools/bin/macos_assemble.sh\n";
+			shellScript = "touch Flutter/ephemeral/tripwire\n\"$FLUTTER_ROOT\"/packages/flutter_tools/bin/macos_assemble.sh\n";
```
stuartmorgan: Touching the tripwire should be last, not first, to ensure that it's newer than all the outputs even if things are created/copied by `assemble`

It would also help to put a comment on the line explaining why it's there and referencing an issue link.

PR #542, opened 2019-09-06
updated vsproj prop for updated repo code

canewsin:
vscode prop stays the same as GLFW Example


stuartmorgan:
Is this value used for anything in a project that doesn't use .NET? Where were you seeing this show up?

If it's not relevant to the project, we could just remove it unless it's required by VS. If it is, it should be something generic like "FlutterRunner".

stuartmorgan:
Closing due to lack of response.

Issue #556, opened 2019-09-12
Running failed


mofada:
**Describe the bug**
run failure.
```
PS F:\WorkSpace\Flutter\flutter-desktop-embedding\example> flutter run -d Windows
Running "flutter pub get" in example...                             1.4s
Launching lib\main.dart on Windows in debug mode...
Building Windows application...
Build process failed. To view the stack trace, please run `flutter run -d windows -v`.
PS F:\WorkSpace\Flutter\flutter-desktop-embedding\example> flutter run -d windows -v
[  +19 ms] executing: [D:\flutter\] git -c log.showSignature=false log -n 1 --pretty=format:%H
[  +55 ms] Exit code 0 from: git -c log.showSignature=false log -n 1 --pretty=format:%H
[        ] 32c946f31bfff6c766229d7b85ef0ae70de2c89a
[        ] executing: [D:\flutter\] git describe --match v*.*.* --first-parent --long --tags
[  +58 ms] Exit code 0 from: git describe --match v*.*.* --first-parent --long --tags
[        ] v1.10.1-36-g32c946f31
[   +6 ms] executing: [D:\flutter\] git rev-parse --abbrev-ref --symbolic @{u}
[  +34 ms] Exit code 0 from: git rev-parse --abbrev-ref --symbolic @{u}
[        ] origin/master
[        ] executing: [D:\flutter\] git ls-remote --get-url origin
[  +36 ms] Exit code 0 from: git ls-remote --get-url origin
[        ] https://github.com/flutter/flutter.git
[  +66 ms] executing: [D:\flutter\] git rev-parse --abbrev-ref HEAD
[  +30 ms] Exit code 0 from: git rev-parse --abbrev-ref HEAD
[        ] master
[  +90 ms] executing: D:\Android\android-sdk\platform-tools\adb.exe devices -l
[  +21 ms] Exit code 0 from: D:\Android\android-sdk\platform-tools\adb.exe devices -l
[        ] List of devices attached
           c1f5972a7d74           device product:santoni model:Redmi_4X device:santoni transport_id:1
[  +24 ms] Artifact Instance of 'FlutterWebSdk' is not required, skipping update.
[   +4 ms] Artifact Instance of 'MacOSEngineArtifacts' is not required, skipping update.
[   +1 ms] Artifact Instance of 'LinuxEngineArtifacts' is not required, skipping update.
[...]

PR #557, opened 2019-09-12
Roll minimum Flutter version forward

stuartmorgan:
This should have been done as part of #553; it would have avoided #556.

Since there have been several other recent desktop-related fixes, this
requires a new enough version to pick them up as well to reduce the
chance of people hitting them.


stuartmorgan:
Looks like I broke testbed on Windows (maybe I didn't do a sufficiently clean build) with the recent change and hadn't noticed. Landing this since that's not a regression, and I'll file a bug to fix testbed tomorrow.

PR #563, opened 2019-09-13
added windows system fonts listener

chunhtai:
This will requires https://github.com/flutter/engine/pull/12276/files for the new embedding api

### DIFF
--- a/example/windows/win32_window.cc
+++ b/example/windows/win32_window.cc
@@ -114,6 +114,11 @@ Win32Window::MessageHandler(HWND hwnd, UINT const message, WPARAM const wparam,
         SetFocus(child_content_);
       }
       return 0;
+
+    // Messages that are directly forwarded to embedding.
+    case WM_FONTCHANGE:
+      SendMessage(child_content_, WM_FONTCHANGE, NULL, NULL);
+      return 0;
   }
 
   return DefWindowProc(window_handle_, message, wparam, lparam);
--- a/testbed/windows/win32_window.cc
+++ b/testbed/windows/win32_window.cc
@@ -114,6 +114,11 @@ Win32Window::MessageHandler(HWND hwnd, UINT const message, WPARAM const wparam,
         SetFocus(child_content_);
       }
       return 0;
+
+    // Messages that are directly forwarded to embedding.
+    case WM_FONTCHANGE:
+      SendMessage(child_content_, WM_FONTCHANGE, NULL, NULL);
+      return 0;
   }
 
   return DefWindowProc(window_handle_, message, wparam, lparam);


chunhtai:
@stuartmorgan This is ready for review. For some reason, I can't add a reviewer

stuartmorgan:
Will try to review tomorrow; currently swamped.

example/windows/win32_window.h
```diff
@@ -40,7 +42,7 @@ class Win32Window {
   // as logical pixels and scale to appropriate for the default monitor. Returns
   // true if the window was created successfully.
   bool CreateAndShow(const std::wstring &title, const Point &origin,
-                     const Size &size);
+                     const Size &size, flutter::FlutterViewController *controller);
```
stuartmorgan: I would like to see a limited-purpose interface (e.g., FlutterWindowDelegate or similar) passed rather than the whole controller, with only the specific methods the window needs access to.

example/windows/win32_window.cc
```diff
@@ -114,6 +116,9 @@ Win32Window::MessageHandler(HWND hwnd, UINT const message, WPARAM const wparam,
         SetFocus(child_content_);
       }
       return 0;
+    case WM_FONTCHANGE:
+      controller_->OnFontChange();
```
stuartmorgan: I don't know enough about Windows to know if this is a reasonable suggestion or not, but I'm curious if we could forward `WM_FONTCHANGE` to `child_content_` and avoid the need for the special-purpose APIs at the embedding layer. Then we could just handle `WM_FONTCHANGE` in the embedding layer (in fact, you already do, where I commented wondering whether that code was doing anything).

@clarkezone Is that doable? A good/bad idea?

clarkezone: Seems reasonable.  I assume WM_FONTCHANGE isn’t automatically propogated to child windows hence the approach here, but manually posting it would avoid the need for the API as you say.

chunhtai: Is there a standard way to forwarding windows message? It does not seems like windows provide an api to reroute message to child process. If we want to use sendMessage to forward the message, the embedding will have to start a message loop as well

stuartmorgan: > reroute message to child process

It's not a child process, it's just a different HWND in the same process.

> If we want to use sendMessage to forward the message, the embedding will have to start a message loop as well

Why would that be the case? The child window already receives messages (e.g., mouse events), so clearly the existing message loop is capable of sending messages to that HWND.

Also, the docs for SendMessage say: "If the specified window was created by the calling thread, the window procedure is called immediately as a subroutine." I would expect this is the same thread.

chunhtai: i made it too complicated, fixed

example/windows/win32_window.cc
```diff
@@ -114,6 +114,11 @@ Win32Window::MessageHandler(HWND hwnd, UINT const message, WPARAM const wparam,
         SetFocus(child_content_);
       }
       return 0;
+
+    // Messages directly forward to embedding.
```
stuartmorgan: Nit: s/forward/forwarded/

stuartmorgan:
You'll need to fix the commit email address and force-push to address the CLA issue, then it'll be good to land.

While you're doing that, could you replicate this change in `testbed/`? I'd like them to stay in sync as much as possible.

chunhtai:
> You'll need to fix the commit email address and force-push to address the CLA issue, then it'll be good to land.
> 
> While you're doing that, could you replicate this change in `testbed/`? I'd like them to stay in sync as much as possible.

will do, for some reason my windows desktop does not play well with github, will fixes it once i finalize both engine embedding and here.

PR #564, opened 2019-09-14
path_provider_fde, url_launcher_fde: Add iOS/Android noop plugins to satisfy gradle and cocoapods

hpoul:
If you are still accepting noop plugins while we are waiting for https://github.com/flutter/flutter/issues/39657 ;-)
i've been adding them for the `path_provider_fde` and `url_launcher_fde` plugins.

thanks


plugins/flutter_plugins/url_launcher_fde/pubspec.yaml
```diff
@@ -7,6 +7,8 @@ homepage: https://github.com/google/flutter-desktop-embedding/tree/master/plugin
 flutter:
   plugin:
     platforms:
+      ios:
+        pluginClass: UrlLauncherFakeFdePlugin
```
stuartmorgan: Was this necessary to fix an error? I would not expect this to be needed (or desirable).

plugins/flutter_plugins/path_provider_fde/ios/Classes/PathProviderPlugin.h
```diff
@@ -0,0 +1,4 @@
+#import <Flutter/Flutter.h>
+
+@interface PathProviderPlugin : NSObject<FlutterPlugin>
```
stuartmorgan: This should be called FDEPathProviderPlugin (and the filename changed accordingly). I wasn't careful about that in the previous reviews, but since avoiding collisions definitely matters here the naming should be done correctly.

plugins/flutter_plugins/url_launcher_fde/ios/Classes/UrlLauncherFakeFdePlugin.m
```diff
@@ -0,0 +1,6 @@
+#import "UrlLauncherFakeFdePlugin.h"
+
+@implementation UrlLauncherFakeFdePlugin
```
stuartmorgan: FDEURLLauncherPlugin

plugins/flutter_plugins/url_launcher_fde/pubspec.yaml
hpoul: Interesting, tbh - I'm not sure, I thought it was.. maybe in an older version? anyway, I removed that and tested it on flutters stable and master channel.. and it also makes the actual classes redundant. removed them..

plugins/flutter_plugins/url_launcher_fde/ios/Classes/UrlLauncherFakeFdePlugin.m
hpoul: completely removed the class, the podspec seems to be enough..

plugins/flutter_plugins/path_provider_fde/ios/Classes/PathProviderPlugin.h
hpoul: removed it

hpoul:
@stuartmorgan thanks for your quick response, seems the classes were redundant, maybe in an old flutter version it was still required? no idea. but it seems to work in master and stable channel now by just having the podspec..

actually it now looks like it would be a very easy fix in the projects `Podfile` to simply check if a `ios` path exists before [including the pod](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/templates/cocoapods/Podfile-ios-objc#L62) .. but this sounds a bit error prone..

stuartmorgan:
> maybe in an old flutter version it was still required?

Prior to the new plugin declaration schema, a registrant entry would have been generated for every plugin unconditionally on iOS.

> actually it now looks like it would be a very easy fix in the projects `Podfile` to simply check if a `ios` path exists

See discussion in 39659.

PR #565, opened 2019-09-15
Apply the toolchain branching to FlutterBuild

stuartmorgan:
Even though FlutterBuild.vcxproj doesn't actually use the C++ toolchain,
VS checks it as part of the build process, so listing only 141 prevents
building in a VS 2019 install that only has v142 (the default).

Fixes https://github.com/flutter/flutter/issues/40532


stuartmorgan:
Landing TBR since the bug will break a bunch of people using VS 2019.

rydmike:
Solved my build issue. Thanks for the quick fix, on a Sunday! Unbelievable and amazing!

PR #574, opened 2019-09-24
add more issue templates

lesnitsky:
@stuartmorgan I'm watching this repo and I see a lot of window/plugin related questions, so I added issue templates which might answer the question before submitting the issue

### DIFF
--- a/.github/ISSUE_TEMPLATE/desktop_window.md
+++ b/.github/ISSUE_TEMPLATE/desktop_window.md
@@ -1,0 +1,14 @@
+---
+name: Window styling/size realted issue
+about: I want to apply custom styles to application window
+title: ""
+labels: ""
+assignees: ""
+---
+
+<!--
+Flutter itself has nothing to do with OS sepcific window styling
+You need to tune everything related to window appearance in macos/windows/linux projects
+-->
+
+**Your question**
--- a/.github/ISSUE_TEMPLATE/plugin_question.md
+++ b/.github/ISSUE_TEMPLATE/plugin_question.md
@@ -1,0 +1,14 @@
+---
+name: Plugin related question
+about: I have a question whether I can use certain plugins in flutter-desktop
+title: ""
+labels: ""
+assignees: ""
+---
+
+<!--
+You can find list of available plugins [here](https://github.com/google/flutter-desktop-embedding/tree/master/plugins)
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE/desktop_window.md
```diff
@@ -0,0 +1,14 @@
...
+name: Window styling/size realted issue
+about: I want to apply custom styles to application window
+title: ""
+labels: ""
+assignees: ""
+---
+
+<!--
+Flutter itself has nothing to do with OS sepcific window styling
+You need to tune everything related to window appearance in macos/windows/linux projects
```
stuartmorgan: This isn't entirely true. I would rather get these issues, and decide which ones should be feature requests for the API surface.

.github/ISSUE_TEMPLATE/plugin_question.md
```diff
@@ -0,0 +1,14 @@
+---
+name: Plugin related question
+about: I have a question whether I can use certain plugins in flutter-desktop
+title: ""
+labels: ""
+assignees: ""
+---
+
+<!--
+You can find list of available plugins [here](https://github.com/google/flutter-desktop-embedding/tree/master/plugins)
```
stuartmorgan: This isn't necessarily true; hopefully macOS plugin implementations will start being available in other places before long. (And theoretically people could host unpublished Windows and Linux plugins as FDE currently does, they would just be subject to breakage so I wouldn't necessary recommend it.)

.github/ISSUE_TEMPLATE/plugin_question.md
```diff
@@ -0,0 +1,14 @@
...
+Note: Most of the flutter plugins are depending on android and ios native SDKs, so most likely you won't be able to use most of flutter plugins as-is
```
stuartmorgan: I think this is misleading.
- First, by definition no existing mobile plugin can be used as-is. E.g., SQLite is available pretty much everywhere, but you still can't use the iOS or Android sqflite plugin implementation on any desktop platform.
- Second, I think it conflates "is this currently available" with "could this be written". We only currently have `path_provider` for macOS, for instance, but someone could write the other platforms at any point, and we in fact have a bug tracking that. That's completely different than questions about Firebase, where the underlying SDK does not itself support desktop.

I don't think there's a good way to answer the general category of questions in a way that's short enough that people will actually read it.

PR #581, opened 2019-09-27
Reconcile Windows example and testbed

stuartmorgan:
The 'example' and 'testbed' projects have diverged over time more than
necessary; this updates them to minimize differences:
- Adds incude and link directories for plugins to example; this makes it
  easier to add plugins to the example, rather than needing to start
  with testbed.
- Adds plugin dll copying to example's bundle_assets_and_deps; again,
  this makes it easier to add plugins to a project that started from
  example.
- Minor cleanup for accidental drift:
  - example wasn't including ProjectDir in the include list
  - Source files were listed in a different order
  - An x86 configuration was accidentally left in testbed's Runner.sln
  - An earlier name for the icon file was left in one project.


stuartmorgan:
(Not sending for review yet as this was written on macOS, so needs to run through CI to find the inevitable mistakes.)

PR #583, opened 2019-09-28
Remove ignored BUILD declaration in Linux Makefile.

mtcliatt:


stuartmorgan:
After having thought about it a bit more I'm actually going to do https://github.com/flutter/flutter/issues/41528 instead, which means this will still be needed as a default.

PR #588, opened 2019-10-09
Update Windows runloop for task API changes

stuartmorgan:
Updates the Runners for https://github.com/flutter/engine/pull/13043, which is a breaking change, and makes the runloop handle engine messages even where there are no pending Win32 messages, using the new information the updated API provides.


example/windows/main.cpp
```diff
@@ -84,9 +85,30 @@ int APIENTRY wWinMain(HINSTANCE instance, HINSTANCE prev, wchar_t *command_line,
...
+    while (PeekMessage(&message, nullptr, 0, 0, PM_REMOVE)) {
+      if (message.message == WM_QUIT) {
+        window.Destroy();
+        break;
+      }
+      TranslateMessage(&message);
+      DispatchMessage(&message);
+    }
+    // Allow Flutter to process its messages.
+    // TODO: Consider interleaving processing on a per-message basis to avoid
```
franciscojma86: Should we file a bug on this?

stuartmorgan: I figured we'd let it bake in the real world for a while to see if it's something that actually comes up in practice. The code gets uglier if we do this, which is why I held off for now.

PR #598, opened 2019-10-21
Windows implementation of the shared_preferences plugin

altera2015:
Here is a Windows back-end implementation for the shared_preferences plugin. It uses the Windows registry to store the values under HKCU\Software\Fluter<EXECUTABLE_NAME>...

### DIFF
--- a/plugins/flutter_plugins/shared_preferences_fde/pubspec.yaml
+++ b/plugins/flutter_plugins/shared_preferences_fde/pubspec.yaml
@@ -9,7 +9,9 @@ flutter:
     platforms:
       macos:
         pluginClass: SharedPreferencesPlugin
-
+      windows:
+        pluginClass: SharedPreferencesPlugin
+        
 environment:
   sdk: ">=2.1.0 <3.0.0"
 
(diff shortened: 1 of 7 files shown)


ignaciotcrespo:
I don't like the idea of polluting the registry with preferences for dozens of apps 🤔

altera2015:
Isn't that what the Windows registry was designed for?

stuartmorgan:
This is a very different approach than what had been discussed--and is already being worked on--in #434. Could you explain why you believe this is a better approach than storing data in AppData?

@csells should weigh in since he's working on that, and has a Windows background, but I just did a bit of reading up on this and it doesn't seem like the registry is recommended for storing arbitrary user data in an application.

altera2015:
One advantage is that it has no dependencies and very little code. But really I have no big opinion either way. This was easy to implement and I needed it sooner rather than later. Feel free to discard if not useful.

csells:
The Registry has come and gone as the recommended spot for user settings on Windows. Also, we're doing a bit of a refactor of this plugin (and probably others) such that you aren't forced to use platform messages when implementing it. That would allow, for example, using Dart FFI or a single platform message to get access to the AppData folder on the machine and then the rest of the preferences saving/loading can all be done in Dart code.

clarkezone:
Agreed, recommend not to user the registry

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

altera2015:
@csells That refactor sounds like a solid plan, too much was pushed to the platform side. Closing this request in anticipation of the new plugin.

PR #615, opened 2019-11-27
Initial macOS implementation of min/max window size

domesticmouse:
PTAL @csells


domesticmouse:
Related bug: https://github.com/google/flutter-desktop-embedding/issues/483

plugins/color_panel/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -0,0 +1,12 @@
+//
```
stuartmorgan: I haven't reviewed the patch in general yet, but as an initial note: this file should not be checked in for any plugin.

domesticmouse: Deleted all of them and added a line to `.gitignore`. Thanks!

plugins/window_size/lib/src/window_size_channel.dart
```diff
@@ -26,31 +26,60 @@ const String _windowSizeChannelName = 'flutter/windowsize';
...
+
 /// The method name to set the frame of a window.
 ///
 /// Takes a frame array, as documented for the value of _frameKey.
 const String _setWindowFrameMethod = 'setWindowFrame';
 
+/// The method name to set the minimum size of a window.
+///
+/// Takes a window size array, as documented for the value of _windowSizeKey.
+const String _setWindowMimumumSizeMethod = 'setWindowMinimumSize';
```
timsneath: Should be `_setWindowMinimumSizeMethod`, not `_setWindowMimumumSizeMethod`, no? Same for `_getWindowMimumumSizeMethod` (as well as references to both).

domesticmouse: Fixed, thanks for the review =)

domesticmouse: Agreed, and fixed. Thanks!

plugins/window_size/lib/src/window_size_channel.dart
```diff
@@ -26,31 +26,60 @@ const String _windowSizeChannelName = 'flutter/windowsize';
...
+
+/// The method name to get the minimum size of a window.
+///
+/// Returns a window size array, as documented for the value of _windowSizeKey.
+const String _getWindowMinimumSizeMethod = 'getWindowMinimumSize';
+
+/// The method name to get the maximum size of a window.
+///
+/// Returns a window size array, as documented for the value of _windowSizeKey.
+const String _getWindowMaximumSizeMethod = 'getWindowMaximumSize';
```
stuartmorgan: I think the method channel needs to clearly define what "no constraint" returns, and how to set it. Currently you're leaking whatever the platform implemention is (e.g., `FLT_MAX` for macOS) which would make removing constraints, or checking if there is a constraint, difficult on the Dart side, requiring platform-specific switching.

Something should be defined for the channel (null argument maybe?) and the platform implementions should translate to the right representation is for the platform.

testbed/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -1,26 +0,0 @@
-//
```
stuartmorgan: This needs to be reverted as well.

.gitignore
```diff
@@ -1 +1,16 @@
...
+.packages
+.pub-cache/
+.pub/
+/build/
+
+# Flutter generated files
+.flutter-plugins-dependencies
+flutter_export_environment.sh
+Flutter-Generated.xcconfig
+GeneratedPluginRegistrant.swift
```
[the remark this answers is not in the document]
domesticmouse: Reverted change

example/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -1,10 +0,0 @@
-//
```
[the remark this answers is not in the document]
domesticmouse: Reverted change

testbed/macos/Flutter/GeneratedPluginRegistrant.swift
domesticmouse: Reverted change

plugins/window_size/lib/src/window_size_channel.dart
domesticmouse: Null feels kinda ick, especially in the days of not null by design. I can't think of anything better, apart from using a string and having a known value for unconstrained.

plugins/window_size/lib/src/window_size_channel.dart
```diff
@@ -121,6 +150,50 @@ class WindowSizeChannel {
     }
   }
 
+  /// Sets the minimum size of the window containing this Flutter instance.
+  void setWindowMinSize(Size size) async {
+    try {
```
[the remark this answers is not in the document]
domesticmouse: I'm not a flutter/plugin specialist, this is the first plugin I've worked on with any seriousness. Is it worth dragging in someone skilled in this area for a best practices review?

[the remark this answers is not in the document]
stuartmorgan: I just checked a few flutter/plugins, and they don't wrap. We should remove all the try/catches here.

plugins/window_size/lib/src/window_size_channel.dart
stuartmorgan: It doesn't need to be a string; we could define `0` to be unconstrained. For min size, that's equivalent to unconstrained anyway, and for max 0 doesn't make sense, so would never come up in normal usage. Similarly, we could use `-1` since we're not dealing with unsigned ints, and `-1` would never be used.

plugins/window_size/lib/src/window_size_channel.dart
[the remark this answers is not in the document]
domesticmouse: Try/catch wrappers removed

plugins/color_panel/.gitignore
```diff
@@ -1,3 +1,8 @@
 .packages
 .flutter-plugins
 pubspec.lock
+.flutter-plugins-dependencies
```
domesticmouse: Hey @stuartmorgan, what are your feelings on adding these entries to the individual `.gitignore` files?

plugins/window_size/lib/src/window_size_channel.dart
domesticmouse: One upside to using `FLT_MAX` as unconstrained sizing for max width and max height is that `Size.fromWidth` and `Size.fromHeight` work out of the box. It looks like Flutter is already using float max as sentinel for unconstrained maximum size.

plugins/color_panel/.gitignore
stuartmorgan: I'd much rather Flutter not spew files into plugin directories.

It'll be irrelevant once the recommended way to consume these plugins isn't by path. Since that's a hack anyway, I'm fine with either compensating with this, or ignoring the problem until it goes away. As noted previously, the latter has been my approach so far.

plugins/color_panel/.gitignore
```diff
@@ -1,3 +1,8 @@
...
+package_config.json
+flutter_export_environment.sh
+Flutter-Generated.xcconfig
+GeneratedPluginRegistrant.swift
```
stuartmorgan: Several of these are macOS-specific, and should go in the macos directory.

plugins/example_plugin/.gitignore
```diff
@@ -1,3 +1,8 @@
 .packages
 .flutter-plugins
 pubspec.lock
+.flutter-plugins-dependencies
+package_config.json
+flutter_export_environment.sh
+Flutter-Generated.xcconfig
+GeneratedPluginRegistrant.swift
```
stuartmorgan: For example_plugin please put anything that deviates from the standard Flutter plugin template rules in a separate section starting with a comment explaining why there are extra ignore rules as compared to the template. Anything checked in to the example will spread to other repos, where we can't clean it up later, so it should be clearly explained.

I would also rather you not touch every plugin in the unrelated PR. If you want to do this in a separate PR that's fine, but this change is unrelated to what this PR is about.

plugins/window_size/lib/src/window_size_channel.dart
stuartmorgan: But it relies on every native implementation having the same definition of float size; there's no real reason that needs to be consistent across all languages. `FLT_MAX` doesn't even have a single defined value in C++ for instance.

I would much rather we pick something clear and explicit, and convert in every implementation, than have something that happens to work right now but isn't guaranteed to have consistent behavior in the future.

domesticmouse: So, how about `double.infinity` in the exposed Dart API for congruence with `Size.fromHeight` and `Size.fromWidth`, and `-1` for across the channel from Dart land to the native code? I like `-1` as it is obviously invalid for all of the sizing properties, and I'm hoping `-1f` has a single defined value in C++.

plugins/color_panel/.gitignore
domesticmouse: Ok I'll take the ignore it until it goes away approach.

plugins/color_panel/.gitignore
domesticmouse: reverted

plugins/example_plugin/.gitignore
domesticmouse: reverted

plugins/window_size/lib/src/window_size_channel.dart
stuartmorgan: > So, how about `double.infinity` in the exposed Dart API for congruence with `Size.fromHeight` and `Size.fromWidth`, and `-1` for across the channel from Dart land to the native code?

Sounds good to me.

> I'm hoping `-1f` has a single defined value in C++.

I'm not sure what you mean exactly; -1 *is* a value. The problem with `FLT_MAX` is that it isn't a value, it’s a concept whose value depends on the size of `float`, which it itself not fixed.

domesticmouse: I've made an initial pass at always using -1 to mark min or max as unconstrained through the channel. It's been long enough since I've done Obj C that I'm questioning my approach with converting the values, and I'd appreciate guidance.

I also ran into some issues with some of the plugins used in the testbed gaining endorsed macos versions, so I converted everything across. My temporary solution is going to break windows and linux, so I'd appreciate guidance here as well.

/cc @franciscojma86

stuartmorgan: The solution to the endorsed plugin issues you're hitting are that we need to remove all the macOS implementation of Flutter plugins (in another PR). We were waiting for all of them to land, which I think has happened now. I can probably do it tomorrow.

stuartmorgan: https://github.com/google/flutter-desktop-embedding/pull/621

stuartmorgan: Looking at the code, it seems like using `0` for unconstrained minimum size would be much simpler, since it's an obviously meaningful value for lower bounds, and doesn't have any value ambiguity like the max values do. We can still use -1 for unconstrained max, but that eliminates a bunch of conversion.

plugins/window_size/lib/src/window_size_channel.dart
```diff
@@ -113,12 +128,82 @@ class WindowSizeChannel {
...
+  /// Sets the minimum size of the window containing this Flutter instance.
+  void setWindowMinSize(Size size) async {
+    await _platformChannel.invokeMethod(_setWindowMinimumSizeMethod, [
+      _setWindowMinSizeUnconstrainedCheck(size.width),
+      _setWindowMinSizeUnconstrainedCheck(size.height),
+    ]);
+  }
+
+  // Window maximum size unconstrained is passed over the channel as -1.
+  double _setWindowMaxSizeUnconstrainedCheck(double size) {
```
stuartmorgan: This method isn't setting anything, and I'm not sure what `check` is supposed to communicate exactly, so this name doesn't seem to me to clearly communicate what the method does. I would call this something like `_channelRepresentationForMaxDimension` (I wouldn't use Size since it doesn't actually take a size)

plugins/window_size/lib/src/window_size_channel.dart
```diff
@@ -113,12 +128,82 @@ class WindowSizeChannel {
...
+    if (size != double.infinity) {
+      return size;
+    }
+    return -1;
```
stuartmorgan: Optional: consider writing this method as `return size == double.infinity ? -1 : size;`

plugins/window_size/lib/src/window_size_channel.dart
```diff
@@ -113,12 +128,82 @@ class WindowSizeChannel {
...
+      List<double>.from(
+        response[_windowSizeKey]
+            .cast<double>()
+            .map(_getWindowMinSizeUnconstrainedCheck),
+      ),
+    );
+  }
+
+  // Window maximum size unconstrained is passed over the channel as -1.
+  double _getWindowMaxSizeUnconstrainedCheck(double size) {
```
stuartmorgan: Essentially the same comment here; `_maxDimensionFromChannelRepresentation` would be clearer IMO.

plugins/window_size/macos/Classes/FLEWindowSizePlugin.m
```diff
@@ -71,6 +76,46 @@ - (NSArray *)platformChannelRepresentationForFrame:(NSRect)frame;
...
+    if (size != -1) {
+      return size;
+    }
+    return 0;
+}
+
+/**
+ * Converts the channel representation for unconstrained maximum size `-1` to Cocoa's specific maximum size of `FLT_MAX`.
+ */
+static double _setWindowMaxSizeUnconstrainedCheck(double size) {
```
stuartmorgan: Same comments about naming (and potentially collapsing using `?:`) here as on the Dart side.

Also, style-wise, No `_`, and the first letter needs to be capitalized.

plugins/window_size/macos/Classes/FLEWindowSizePlugin.m
```diff
@@ -71,6 +76,46 @@ - (NSArray *)platformChannelRepresentationForFrame:(NSRect)frame;
...
+    if (size != -1) {
```
stuartmorgan: For clarity, please use a double constant (`-1.0`) rather than integer.

testbed/lib/main.dart
```diff
@@ -323,17 +348,13 @@ class FileChooserTestWidget extends StatelessWidget {
...
-                    allowsMultipleSelection: true,
-                    initialDirectory: initialDirectory)
-                .then(
-              (result) {
-                Scaffold.of(context).showSnackBar(SnackBar(
-                    content: Text(_resultTextForFileChooserOperation(
-                        _FileChooserType.open, result))));
-              },
-            );
+            final result = await file_chooser.showOpenPanel(
```
stuartmorgan: I'm happy to take improvements to the Testbed code in general, but this (and the change above) is unrelated to the PR. I would much rather you just send a PR containing just the miscellaneous cleanup, which could be trivially reviewed and landed separately from the behavior-altering change that this PR is intended to be making.

stuartmorgan:
If you merge the latest master, the plugin build errors should be gone.

domesticmouse:
I’ll get to the fix ups when I stop bouncing between cities. Early next week, hopefully.

plugins/window_size/lib/src/window_size_channel.dart
```diff
@@ -73,35 +102,21 @@ class WindowSizeChannel {
...
-    try {
-      final screenList = <Screen>[];
-      final response =
-          await _platformChannel.invokeMethod(_getScreenListMethod);
-
-      for (final screenInfo in response) {
-        screenList.add(_screenFromInfoMap(screenInfo));
-      }
-      return screenList;
-    } on PlatformException catch (e) {
```
franciscojma86: Why do you need to remove the try catch? Is there no possibility for platform exceptions anymore?

testbed/lib/main.dart
```diff
@@ -62,6 +67,26 @@ void main() {
   runApp(new MyApp());
 }
 
+Future<void> _windowMinMax() async {
```
franciscojma86: This method name isn't very clear on what it does.

plugins/window_size/lib/src/window_size_channel.dart
stuartmorgan: See my comments earlier in the review; I don't think we ever should have been suppressing these exceptions.

testbed/lib/main.dart
domesticmouse: Reverting. Due to bad life choices (creating this PR in the master branch of my fork), these fix ups will have to wait until a follow up PR.

plugins/window_size/lib/src/window_size_channel.dart
domesticmouse: Now using 0 for unconstrained minimum size.

testbed/lib/main.dart
domesticmouse: It was testing the channel communication for min/max value conversion. No longer required.

domesticmouse:
PTAL @stuartmorgan

plugins/window_size/lib/src/window_size_channel.dart
stuartmorgan: The special values for min and max should be documented in these comments.

plugins/window_size/lib/src/window_size_channel.dart
```diff
@@ -113,12 +128,53 @@ class WindowSizeChannel {
...
+      _channelRepresentationForMaxDimension(size.height),
+    ]);
+  }
+
+  /// Gets the minimum size of the window containing this Flutter instance.
+  Future<Size> getWindowMinSize() async {
+    final response =
+        await _platformChannel.invokeMethod(_getWindowMinimumSizeMethod);
+    return _sizeFromWHList(
+        List<double>.from(response[_windowSizeKey].cast<double>()));
```
stuartmorgan: If there's only one entry then there's no need for a map, and this can just return the list directly (and I'm not seeing any reason there would be anything more in the future, nor does the setter use this extra layer). Then you can use  [invokeListMethod](https://api.flutter.dev/flutter/services/MethodChannel/invokeListMethod.html) to avoid the explicit casting here as well.

Same for max size.

plugins/window_size/lib/src/window_size_channel.dart
domesticmouse: Done

plugins/window_size/lib/src/window_size_channel.dart
domesticmouse: Done

domesticmouse:
PTAL @stuartmorgan

domesticmouse:
Thanks for leading the way for me on this PR Stuart =)

PR #620, opened 2019-12-06
Covert file_chooser API to Futures

### LINKED ISSUE #617: Convert file_chooser API to Futures
The current callback-based API is essentially an ObjC API translated to Dart. It would be more idiomatic to return a result Future.

stuartmorgan:
file_chooser was written with an ObjC-like callback API, when an API
based on futures would be more idiomatic for Dart. This converts to a
Futures-based API.

Also changes the implementation so that missing plugin exceptions will
be propagated to clients for handling, rather than just logged.

This is a breaking change for clients of the file_chooser plugin.

Fixes #617


plugins/file_chooser/lib/src/channel_controller.dart
```diff
@@ -116,26 +121,18 @@ class FileChooserChannelController {
...
+  /// Shows a file chooser of [type] configured with [options], returning a
+  /// [FileChooserResult] when complete.
+  Future<FileChooserResult> show(
+    FileChooserType type,
+    FileChooserConfigurationOptions options,
+  ) async {
+    final methodName = type == FileChooserType.open
+        ? _kShowOpenPanelMethod
+        : _kShowSavePanelMethod;
+    final response = await _channel.invokeMethod(
```
jonahwilliams: Sorry, random drive-by comment, use [MethodChannel.invokeListMethod](https://api.flutter.dev/flutter/services/MethodChannel/invokeListMethod.html)

```suggestion
    final paths = await _channel.invokeListMethod<String>(
```

stuartmorgan:
@krisgiesing in testbed/ it's just changes to conform, but the plugins/ part of this PR *is* the API change. Does that API look good for your use? (Also, not eating exceptions is a logic change.)

plugins/file_chooser/lib/src/channel_controller.dart
stuartmorgan: TIL! Thanks for that drive-by :)

krisgiesing:
@stuartmorgan the API changes look good. I'm happy about not eating exceptions.

PR #622, opened 2019-12-11
Update to Swift 5

stuartmorgan:


stuartmorgan:
Apparently I never landed this, and still being on 4 seems to be causing CI issues.

Issue #105, opened 2018-07-12
[windows] Implement existing FDE plugins


stuartmorgan:
This is general bug to track adding Windows support to the existing plugins within this project. If any of them require significant discussion, we can split out individual bugs at the time someone decides to work on one of them.

The list (which I'll try to keep updated over time) is:

- [ ] ~color_panel~
- [x] file_chooser
- [x] menubar
- [x] window_resize

~This is obviously blocked on #103.~

axellebot:
Hello,

I'm kind of lost between all those repositories.

- Where should we add/update plugins for desktop ?
https://github.com/google/flutter-desktop-embedding OR https://github.com/flutter/plugins

- Is there any template to create a plugin for windows or enhance a existing plugin ?

FYI I'd like to add windows feature to the `shared_preferences` from https://github.com/google/flutter-desktop-embedding

stuartmorgan:
> FYI I'd like to add windows feature to the `shared_preferences` from https://github.com/google/flutter-desktop-embedding

@axellebot If this is something you're still interested in, the road to doing so is now much more clearly paved. `/plugins/flutter_plugins/` now exists to hold such implementations (with #434 tracking shared_preferences specifically), and `/plugins/example_plugin` would serve as a template.

axellebot:
@stuartmorgan Thanks for all your answers, I'll dig into this later this summer.

aloisdeniel:
For the Windows implementation, as soon as the C# plugin is ready, a lot could be taken from the [Xamarin.Essentials](https://github.com/xamarin/Essentials) implementations.

stuartmorgan:
None of these plugins is particularly complex. I would strongly prefer not to have implementations that are based on third-party code, as that adds a lot of complexity to the licensing situation.

stuartmorgan:
window_size will be done shortly. Implementing it turned up some missing plugin registrar functionality that was lost during the Win32 conversion, but once that's in place it'll be straightforward.

stuartmorgan:
I suspect menubar is going to run into the same need for a system for delegating window messages to plugins that has come up with min/max window size and cursor handling.

stuartmorgan:
With `menubar` landed and `color_panel` removed from the repository entirely, there's now a Windows implementation of all of the FDE plugins.

csells:
wahoo!

PR #635, opened 2020-01-10
Implement window_size for Windows

stuartmorgan:
Adds a Windows implementation of the window_size plugin for Windows.
Like the Windows runners, it currently requires Windows 10, so will need
fallback logic in the future if we add fallback code to the runners.

Updates testbed to use the plugin on Windows as well.

Part of #105


stuartmorgan:
Depends on https://github.com/flutter/engine/pull/15378, so I'll need to wait until that's landed and rolled into the framework before landing this.

plugins/window_size/windows/window_size_plugin.cpp
```diff
@@ -0,0 +1,195 @@
...
+EncodableValue GetPlatformChannelRepresentationForMonitor(HMONITOR monitor) {
+  if (!monitor) {
+    return EncodableValue();
+  }
+
+  MONITORINFO info;
+  info.cbSize = sizeof(MONITORINFO);
+  GetMonitorInfo(monitor, &info);
+  UINT dpi_x, dpi_y;
+  GetDpiForMonitor(monitor, MDT_EFFECTIVE_DPI, &dpi_x, &dpi_y);
```
franciscojma86: From https://docs.microsoft.com/en-us/windows/win32/api/shellscalingapi/nf-shellscalingapi-getdpiformonitor#remarks, this api is not DPI aware. We should check for the awareness level and use  GetDpiForWindow instead. We would need a way to get the DPI per awareness setting. I started something like this here https://github.com/flutter/engine/pull/12735/files#diff-61a7e1c427ae9b307bd89eb9b20229a2R30. I can take it over again and land it before this PR.

stuartmorgan: Yes, I saw that but was confused since it's what the Runner code is currently using, and that's clearly reacting to DPI. And this gives me expected values on my laptop. And we can't use GetDpiForWindows because this is monitor-level information—we don't have a window, and there may not even be one (unless there's some kind of meta-window on each monitor automatically?)

I figured being no worse than the Runner currently is seemed like a reasonable starting place, and then we can revisit this when we've cleaned up the Runner and engine to do DPI coherently.

franciscojma86: Sounds good, we should revisit dpi awareness soon anyway

PR #637, opened 2020-01-10
Support Change Cursor (Windows)

rodydavis:
Need to call the WM_SETCURSOR event and return true if you change the cursor in a plugin.

Fixes: https://github.com/google/flutter-desktop-embedding/issues/636

Supports: https://github.com/rive-app/window-utils/issues/8

### DIFF
--- a/example/windows/win32_window.cc
+++ b/example/windows/win32_window.cc
@@ -95,6 +95,9 @@ Win32Window::MessageHandler(HWND hwnd, UINT const message, WPARAM const wparam,
   }
 
   switch (message) {
+    case WM_SETCURSOR:
+      return TRUE;
+      
     case WM_DESTROY:
       window_handle_ = nullptr;
       Destroy();
--- a/testbed/windows/win32_window.cc
+++ b/testbed/windows/win32_window.cc
@@ -95,6 +95,9 @@ Win32Window::MessageHandler(HWND hwnd, UINT const message, WPARAM const wparam,
   }
 
   switch (message) {
+    case WM_SETCURSOR:
+      return TRUE;
+      
     case WM_DESTROY:
       window_handle_ = nullptr;
       Destroy();


rodydavis:
@googlebot I signed it!

stuartmorgan:
It seems like there are two significant issues here, unless I'm misunderstanding:
- You're unconditionally disabling all default cursor behavior. Unless someone is using your plugin, cursors won't do the right thing. (Also, I suspect that you may see edge case issues even with it if you're not calling your plugin on every mouse move event, with things like stuck cursors from other apps, although I don't know enough about OS-level cursor handling on Windows to be sure.)
- You're changing the cursor behavior for the top-level HWND, rather than for the Flutter HWND, so if someone made the Flutter HWND smaller than the full window, unexpected things would happen.

Someone could certainly make this change locally to their own runner if they know that these things don't apply, but that wouldn't be the right behavior for all cases.

rodydavis:
This should only effect the application lever cursor not the system one. I was careful not to use the system level apis. The default cursor for the system still works when you run the example

stuartmorgan:
> This should only effect the application lever cursor not the system one.

I understand that, but if you unconditionally respond `TRUE` even when you haven't set the cursor, it's not clear to me what happens if, for instance, you move the mouse directly from a window that has set a custom cursor (in another application, potentially) into the Flutter view, but don't actually set the cursor. I wouldn't be at all surprised if that case, or other similar cases, give you a stuck cursor within the Flutter view until your plugin is called again.

stuartmorgan:
(And to clarify: when I say top-level HWND, I'm talking about the thing that the user perceives as a window, has window controls, etc. and lives in FDE's Runner, and when I say Flutter HWND I'm talking about the window/view that is drawing Flutter content, and lives in the embedding implementation. They aren't the same HWND, and Flutter cursor behavior should be constrained to the latter, not the former.)

rodydavis:
Oh gotcha, I see what you are saying. From what I have been reading though there is a system level cursor and application specific one. When you cross bounds into the application it is up to the application to handle the event:

event setting the WNDCLASS cursor to null does not work

```
WNDCLASS window_class{};
// window_class.hCursor = LoadCursor(nullptr, IDC_ARROW);
window_class.hCursor = nullptr;
```

Something else in the embedding is consuming the event.

When you return TRUE to the WM_SETCURSOR event then it will approve setting the cursor. It could be an opt in method for setting a cursor but personally I think the benefits out weigh the cons and edge cases. (Since it is not overriding the system)

rodydavis:
Also the WM_SETCURSOR is only called for your application when the mouse moves.

https://docs.microsoft.com/en-us/windows/win32/menurc/using-cursors

stuartmorgan:
> When you return TRUE to the WM_SETCURSOR event then it will approve setting the cursor.

According to the documentation, when you return `TRUE` it prevents the call from continuing up the chain and eventually calling the default behavior of setting the class cursor, which is very different. It's not like this is a no-op change for people who aren't setting the cursor via your plugin.

You want to return `TRUE` in your case not to "approve" the change you made, but to prevent the default behavior from immediately overriding it with the class cursor again.

> When you cross bounds into the application it is up to the application to handle the event

Yes, and the default behavior is that when the event bubbles all the way up unhandled it will set the cursor to the class cursor. You are unconditionally turning off that behavior, so in an application that doesn't use your plugin nothing in the application will handle it, which is the wrong behavior.

> personally I think the benefits out weigh the cons

For having this as an option, or landing this as-is? For the latter, I strongly disagree, since it makes every application that isn't using your plugin incorrect.

If the former, as I said above if you think it's good enough for your use case you can certainly modify your runner accordingly, as can anyone else interested in using this plugin until the embedding support is ready.

> Also the WM_SETCURSOR is only called for your application when the mouse moves.

I'm aware of that; nothing about the discussion above assumed otherwise.

stuartmorgan:
I'm going to close this, since per my comments above this project doesn't own the `HWND` that should actually be handling `WM_SETCURSOR`. I'm open to reviewing a proposal for a system that allows plugins to register as, essentially, window message delegates of the Flutter view HWND (which would let your plugin insert this behavior at the correct level, and conditionally on the plugin being used), but that would be in flutter/engine rather than here.

rodydavis:
Sounds good!

PR #647, opened 2020-01-21
File file_chooser macOS view reference

### LINKED ISSUE #633: File picker dialog does not block interaction with the Flutter app underneath
**Describe the bug**
My desktop app is using the file_chooser plugin. On a Mac desktop, when I bring up a native dialog via that plugin (eg. showOpenPanel, showSavePanel), my expectation is that until I closed that dialog, I would not be able to interact with the rest of the app. What I'm seeing currently is that I am able to click back into the main app window and interact freely, including both mouse and keyboard events, while the file chooser dialog is still open.

**Doctor Output**
```
[✓] Flutter (Channel unknown, v1.13.3-pre.9, on Mac OS X 10.14.6 18G2022, locale en-US)
[✗] Android toolchain - develop for Android devices
    ✗ Unable to locate Android SDK.
      Install Android Studio from: https://developer.android.com/studio/index.html
      On first launch it will assist you in installing the Android SDK components.
      (or visit https://flutter.dev/setup/#android-setup for detailed instructions).
      If the Android SDK has been installed to a custom location, set ANDROID_HOME to that location.
      You may also want to add it to your PATH environment variable.

[✓] Xcode - develop for iOS and macOS (Xcode 11.1)
[!] Android Studio (not installed)
[✓] IntelliJ IDEA Community Edition (version 2019.1.2)
[✓] Connected device (1 available)
```

stuartmorgan:
The registrar is not guaranteed to have a view at registration time, and
in practice no longer does causing file_chooser to have a nil view. It
should instead be storing the registrar, and getting the view on demand.

Fixes #633

### DIFF
--- a/plugins/file_chooser/macos/Classes/FLEFileChooserPlugin.m
+++ b/plugins/file_chooser/macos/Classes/FLEFileChooserPlugin.m
@@ -28,14 +28,14 @@ static NSString *const kAllowsMultipleSelectionKey = @"allowsMultipleSelection";
 static NSString *const kCanChooseDirectoriesKey = @"canChooseDirectories";
 
 @implementation FLEFileChooserPlugin {
-  // The view displaying Flutter content.
-  NSView *_flutterView;
+  // The plugin registrar, for obtaining the view.
+  NSObject<FlutterPluginRegistrar> *_registrar;
 }
 
-- (instancetype)initWithView:(NSView *)view {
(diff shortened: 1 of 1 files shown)


krisgiesing:
Out of curiosity, do we know what changed to cause the registrar not to have a view at registration time? (I'm pretty sure this is a recent regression.)

stuartmorgan:
Probably https://github.com/flutter/engine/pull/9750

krisgiesing:
Oh, huh, that's much longer ago than I thought. Thanks for fixing :)

PR #654, opened 2020-01-28
Overhaul example plugin, and rename it 'sample'

### LINKED ISSUE #648: Improve example_plugin
There are some things that should be changed in example_plugin to make it easier to use as a sample that I find myself changing each time I copy it for a new plugin:
- The naming scheme is inconsistent with other plugins, which use `foo` as the plugin name (and project name for Windows) and `foo_plugin.*` for the source files. It should be named something that doesn't end in _plugin (possible "sample" to differentiate it from /example, or maybe "plugin_example" for clarity even though that would make for the strange class name `PluginExamplePlugin`).
  - This affects the use of the variable in the PluginInfo.props on Windows
  - It also makes the Makefile usage of example_plugin weird and inconsistent
- The plugins should use the platform-specific registrars, not the base C++ registrar.
- What, if anything, is passed to the constructor should be revisited. Maybe add the registrar, and comment both that and channel about removing it if unnecessary. Maybe only pass the registrar by default (although the pointer gymnastics for the channel are bit confusing for someone with limited C++ experience, so probably better to leave it).

stuartmorgan:
Renames the plugin from 'example_plugin' to 'sample' so that the naming is
consistent with all the actual plugins (which have 'plugin' in the class name,
but not the actual plugin name).

Also makes various improvements to the sample code and documentation,
as described in the associated bug.

Fixes #648


plugins/sample/README.md
```diff
@@ -0,0 +1,41 @@
+# sample_plugin
```
stuartmorgan: This file is showing up as a new file rather than a modification of the old README because it both moved, and changed substantially due to moving a bunch of the specific instructions into comments in the code.

plugins/sample/windows/sample.vcxproj
```diff
@@ -14,7 +14,7 @@
     <VCProjectVersion>15.0</VCProjectVersion>
     <ProjectGuid>{95E21B2C-C18A-4CED-8509-585CB2570FDE}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
-    <RootNamespace>example_plugin</RootNamespace>
+    <RootNamespace>dev_flutter_plugin</RootNamespace>
```
stuartmorgan: Since this isn't currently used by anything, I just gave it a generic value instead of the plugin name, since I think it's likely to not get updated by people, in practice, so long as it's unused and not surfaced anywhere.

plugins/sample/windows/sample_plugin.cpp
```diff
@@ -1,101 +1,109 @@
...
-
-void ExamplePluginRegisterWithRegistrar(
-    FlutterDesktopPluginRegistrarRef registrar) {
-  // The plugin registrar owns the plugin, registered callbacks, etc., so must
-  // remain valid for the life of the application.
-  static auto *plugin_registrar = new flutter::PluginRegistrar(registrar);
-
-  ExamplePlugin::RegisterWithRegistrar(plugin_registrar);
-}
+// Copyright 2019 Google LLC
```
stuartmorgan: This file was using Windows line endings for some reason, unlike all our other code files, thus the everything-changed diff.

plugins/sample/README.md
```diff
@@ -0,0 +1,41 @@
...
+- To create an entirely new plugin, or a federated Windows/Linux implementation
+  of an existing plugin, copy the entire `sample_plugin` folder.
+  - Update the `pubspec.yaml` as normal for a plugin. Look for comments
+    containing `***` and updated those entries as described in the comment.
+  - Delete the directories for any platforms you aren't supporting.
+- To add desktop support directly to an existing plugin, copy the `windows`
+  and/or `linux` directories into that plugin.
+  - Update the plugin's `pubspec.yaml` to include the new platforms.
+    If you haven't already, you'll need to switch to the new `plugins:`
+    format, at the legacy declaration doesn't support desktop.
```
franciscojma86: s/at/as

plugins/sample/README.md
```diff
@@ -0,0 +1,41 @@
...
+or Linux.
+
+Before continuing, read [the main plugins README](../README.md) if you haven't already.
+
+## Use
+
+- To create an entirely new plugin, or a federated Windows/Linux implementation
+  of an existing plugin, copy the entire `sample_plugin` folder.
+  - Update the `pubspec.yaml` as normal for a plugin. Look for comments
+    containing `***` and updated those entries as described in the comment.
```
franciscojma86: s/updated/update

plugins/sample/README.md
stuartmorgan: Done.

plugins/sample/README.md
stuartmorgan: Done.

Issue #343, opened 2019-04-22
Add better warning settings to example projects


stuartmorgan:
As part of evolving those projects into `create` templates, it would be good to add fairly strict warning settings (what we have now is mostly toolchain defaults). It'll be easy to relax them over time if necessary, but hard to make them stricter once they're in general use.

stuartmorgan:
Ref https://github.com/flutter/flutter/issues/32416

stuartmorgan:
Filed https://github.com/flutter/flutter/issues/53172 for remaining work (Linux)

PR #663, opened 2020-02-06
Misc. Windows code cleanup

### LINKED ISSUE #582: [windows] Fix plugin exception warnings
Since the change to turn off exceptions in the plugins and runners (to simplify using /clr), the build shows warnings like:
```
[ +732 ms] c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.16.27023\include\xlocale(319): warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc [C:\projects\flutter-desktop-embedding\plugins\example_plugin\windows\example_plugin.vcxproj]
```

We should figure out what is using exception handlers, and remove them (or if we can't for some reason, re-evaluate that change).

### LINKED ISSUE #662: Clean up Windows style
The Windows code hasn't been consistent about following Google's style recommendations on Windows code. Notably:
- We shouldn't ever be using Hungarian notation for variable names.
- Pointer types shouldn't use the composite Windows types (e.g., `LP...`).

stuartmorgan:
This is a complete pass over all the code and projects, making a variety of changes:
- Updating style to better match Google style for Windows code (e.g., avoiding LP* types)
- Increasing the warning level for everything from W3 to W4
- Enabling warnings as errors (WX)
- Fixing all warnings
  - Including correctly disabling exceptions in STL via a Define
- clang-formatted some missed files

Fixes #662 
Fixes #582 
Addresses Windows portion of #343

### DIFF
--- a/testbed/windows/win32_window.cc
+++ b/testbed/windows/win32_window.cc
@@ -146,7 +146,7 @@ Win32Window *Win32Window::GetThisFromHandle(HWND const window) noexcept {
 
 void Win32Window::SetChildContent(HWND content) {
   child_content_ = content;
-  auto res = SetParent(content, window_handle_);
+  SetParent(content, window_handle_);
   RECT frame;
   GetClientRect(window_handle_, &frame);
 
--- a/testbed/windows/win32_window.cc
+++ b/testbed/windows/win32_window.cc
@@ -13,7 +13,7 @@ namespace {
 // constant for machines running at 100% scaling.
 constexpr int kBaseDpi = 96;
 
-constexpr LPCWSTR kClassName = L"CLASSNAME";
+constexpr const wchar_t kClassName[] = L"CLASSNAME";
 
 // Scale helper to convert logical scaler values to physical using passed in
 // scale factor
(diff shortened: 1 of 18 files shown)


testbed/windows/win32_window.cc
```diff
@@ -13,14 +13,24 @@ namespace {
...
 // scale factor
 int Scale(int source, double scale_factor) {
   return static_cast<int>(source * scale_factor);
 }
 
+// Returns the DPI for the monitor containing, or closest to, |point|.
+UINT GetDpiForMonitorAtPoint(const Win32Window::Point &point) {
+  const POINT target_point = {static_cast<LONG>(point.x),
+                              static_cast<LONG>(point.y)};
+  HMONITOR monitor = MonitorFromPoint(target_point, MONITOR_DEFAULTTONEAREST);
```
stuartmorgan: The reason for this change is that when I was fixing warnings I discovered that passing a null `HWND` to `MonitorFromWindow` is actually a violation of the annotations on the method. The recommended way to get the primary monitor is to call this with a 0,0 point rather than using MonitorFromWindow with null. We should fix that in the embedding.

stuartmorgan:
Looks like I didn't correctly disable the unused parameter warning; I'm not sure why it worked locally. I'll have to fix that next week.

PR #674, opened 2020-02-14
Added linux support for path provider

TimWhiting:
Added Linux support for path provider.
Currently using GNOME paths for application documents directory / applications support directory.
Don't know if that is the best long term solution.
https://askubuntu.com/questions/14535/whats-the-local-folder-for-in-my-home-directory

### DIFF
--- a/plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
+++ b/plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
@@ -1,0 +1,161 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#include "path_provider_plugin.h"
+
+#include <flutter/method_channel.h>
+#include <flutter/plugin_registrar.h>
+#include <flutter/standard_method_codec.h>
+
+#include <fcntl.h>
+#include <linux/limits.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <codecvt>
+#include <cstring>
+#include <iostream>
+#include <locale>
+#include <memory>
+#include <sstream>
+#include <string>
+namespace {
+
+using flutter::EncodableValue;
+
+// Look here for GNOME based directories:
+// https://askubuntu.com/questions/14535/whats-the-local-folder-for-in-my-home-directory
+
+// Returns the name of the executable, without the .exe extension, or an empty
+// string on failure.
+std::string GetExecutableName() {
+  const size_t bufSize = PATH_MAX + 1;
+  char dirNameBuffer[bufSize];
+  // Read the symbolic link '/proc/self/exe'.
+  const char *linkName = "/proc/self/exe";
+  const int ret = int(readlink(linkName, dirNameBuffer, bufSize - 1));
+  if (ret < 0) {
+    return "";
+  }
+  std::string dirName = std::string(dirNameBuffer);
+  return dirName.substr(dirName.find_last_of("/") + 1);
+}
+
+void EnsureCreated(std::string path) {
+  if (mkdir(path.c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH) == -1) {
+    if (errno == EEXIST) {
+      // alredy exists
+    } else {
+      // something else
+      std::cerr << "cannot create " << path << " error:" << strerror(errno)
+                << std::endl;
+    }
+  }
+}
+
+class PathProviderPlugin : public flutter::Plugin {
+ public:
+  static void RegisterWithRegistrar(flutter::PluginRegistrar *registrar);
+
+  virtual ~PathProviderPlugin();
+
+ private:
+  PathProviderPlugin();
+  std::string exe_name;
+  std::string homedir_name;
+  // Called when a method is called on plugin channel;
+  void HandleMethodCall(
+      const flutter::MethodCall<EncodableValue> &method_call,
+      std::unique_ptr<flutter::MethodResult<EncodableValue>> result);
+};
+
+// static
+void PathProviderPlugin::RegisterWithRegistrar(
+    flutter::PluginRegistrar *registrar) {
+  auto channel = std::make_unique<flutter::MethodChannel<EncodableValue>>(
+      registrar->messenger(), "plugins.flutter.io/path_provider",
+      &flutter::StandardMethodCodec::GetInstance());
+
+  // Uses new instead of make_unique due to private constructor.
+  std::unique_ptr<PathProviderPlugin> plugin(new PathProviderPlugin());
+
+  channel->SetMethodCallHandler(
+      [plugin_pointer = plugin.get()](const auto &call, auto result) {
+        plugin_pointer->HandleMethodCall(call, std::move(result));
+      });
+
+  registrar->AddPlugin(std::move(plugin));
+}
+
+PathProviderPlugin::PathProviderPlugin() {
+  exe_name = GetExecutableName();
+  const char *homedir;
+  homedir = getenv("HOME");
+  homedir_name = std::string(homedir);
+}
+
+PathProviderPlugin::~PathProviderPlugin() = default;
+
+void PathProviderPlugin::HandleMethodCall(
+    const flutter::MethodCall<EncodableValue> &method_call,
+    std::unique_ptr<flutter::MethodResult<EncodableValue>> result) {
+  if (method_call.method_name().compare("getTemporaryDirectory") == 0) {
+    std::string result_path = "/tmp";
+    flutter::EncodableValue response(result_path);
+    result->Success(&response);
+  } else if (method_call.method_name().compare(
+                 "getApplicationSupportDirectory") == 0) {
+    if (homedir_name.empty()) {
+      result->Error("Unable to get usr home directory");
+      return;
+    }
+    // Use the executable name as the subdirectory for now.
+    if (exe_name.empty()) {
+      result->Error("Unable to get exe name");
+      return;
+    }
+    std::ostringstream response_stream;
+    response_stream << homedir_name << "/.cache/" << exe_name;
+    EnsureCreated(response_stream.str());
+    flutter::EncodableValue response(response_stream.str());
+    result->Success(&response);
+  } else if (method_call.method_name().compare(
+                 "getApplicationDocumentsDirectory") == 0) {
+    if (homedir_name.empty()) {
+      result->Error("Unable to get usr home directory");
+      return;
+    }
+    // Use the executable name as the subdirectory for now.
+    if (exe_name.empty()) {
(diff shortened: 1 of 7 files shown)


TimWhiting:
@googlebot I signed it!

plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
```diff
@@ -0,0 +1,161 @@
...
+  return dirName.substr(dirName.find_last_of("/") + 1);
+}
+
+void EnsureCreated(std::string path) {
+  if (mkdir(path.c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH) == -1) {
+    if (errno == EEXIST) {
+      // alredy exists
+    } else {
+      // something else
+      std::cerr << "cannot create " << path << " error:" << strerror(errno)
```
stuartmorgan: This function isn't really ensuring that the folder exist, despite the name, since it can fail silently (from an API perspective). If the API contract for the plugin is that the returned directory must exist, then failure should be propagated as an `Error(...)` call with details.

plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
```diff
@@ -0,0 +1,161 @@
...
+
+class PathProviderPlugin : public flutter::Plugin {
+ public:
+  static void RegisterWithRegistrar(flutter::PluginRegistrar *registrar);
+
+  virtual ~PathProviderPlugin();
+
+ private:
+  PathProviderPlugin();
+  std::string exe_name;
```
stuartmorgan: These need trailing `_`, per style guide. Also, "exe" is a Windows-ism; "executable" or "binary" would be more generic.

Although at a higher level: do you actually need to cache these? How expensive are they to look up?

plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
```diff
@@ -0,0 +1,161 @@
...
+  std::string homedir_name;
```
stuartmorgan: This is actually a path, not a name. This variable name makes me think it's the basename.

plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
```diff
@@ -0,0 +1,161 @@
...
+        plugin_pointer->HandleMethodCall(call, std::move(result));
+      });
+
+  registrar->AddPlugin(std::move(plugin));
+}
+
+PathProviderPlugin::PathProviderPlugin() {
+  exe_name = GetExecutableName();
+  const char *homedir;
+  homedir = getenv("HOME");
```
stuartmorgan: If you do keep the caching (although it seems very unlikely that `getenv` is expensive), there's no need for this intermediate variable. `homedir_name = getenv("HOME")` is a much shorter way to write these three lines.

plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
```diff
@@ -0,0 +1,161 @@
...
+}
+
+PathProviderPlugin::~PathProviderPlugin() = default;
+
+void PathProviderPlugin::HandleMethodCall(
+    const flutter::MethodCall<EncodableValue> &method_call,
+    std::unique_ptr<flutter::MethodResult<EncodableValue>> result) {
+  if (method_call.method_name().compare("getTemporaryDirectory") == 0) {
+    std::string result_path = "/tmp";
+    flutter::EncodableValue response(result_path);
```
stuartmorgan: `flutter::EncodableValue response("/tmp");`

plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
```diff
@@ -0,0 +1,161 @@
...
+      result->Error("Unable to get usr home directory");
+      return;
+    }
+    // Use the executable name as the subdirectory for now.
+    if (exe_name.empty()) {
+      result->Error("Unable to get exe name");
+      return;
+    }
+    std::ostringstream response_stream;
+    response_stream << homedir_name << "/.cache/" << exe_name;
```
stuartmorgan: `.cache` doesn't have the right semantics for this directory; a program should be able to store things like, say, an internal database, in this directory, without expecting that it will be removed without warning.

I think based on the docs that you linked to, `.local/share/` may actually be the better match here.

plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
```diff
@@ -0,0 +1,161 @@
...
+      result->Error("Unable to get usr home directory");
+      return;
+    }
+    // Use the executable name as the subdirectory for now.
+    if (exe_name.empty()) {
+      result->Error("Unable to get exe name");
+      return;
+    }
+    std::ostringstream response_stream;
+    response_stream << homedir_name << "/.local/share/" << exe_name;
```
stuartmorgan: The primary distinction between `getApplicationDocumentsDirectory` and the other application data directories in `path_provider` is that this one should be user visible, so a dot directory doesn't seem like the right semantics. The way I've interpreted this on the other desktop platform is to use the general user documents directory. It looks like in XDG terms that would be XDG_DOCUMENTS_DIR.

Also, I haven't been making application-specific subdirs for this one on other desktop platforms. What the right behavior here is isn't obvious, since the API was designed for mobile where everything is sandboxed at the application level in most cases anyway, but for now the approach I've been taking is to just return the documents directory as-is. I'd like to be consistent, and if there's feedback that that's unexpected, change all of the desktop platforms together.

testbed/lib/main.dart
```diff
@@ -321,7 +321,7 @@ class FileChooserTestWidget extends StatelessWidget {
           child: const Text('OPEN'),
           onPressed: () async {
             String initialDirectory;
-            if (Platform.isMacOS || Platform.isWindows) {
+            if (Platform.isMacOS || Platform.isWindows || Platform.isLinux) {
```
stuartmorgan: These are the only platforms this repository is concerned with, so you can just remove the condition.

testbed/linux/Makefile
```diff
@@ -30,6 +30,7 @@ PLUGINS=\
 	menubar \
 	sample \
 	url_launcher_fde \
+	path_provider_fde \
```
stuartmorgan: Nit: This and the list below are sorted alphabetically to make them easier to manage; please preserve that.

plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
stuartmorgan: From what I read starting from your link, it sounds like the expectation is that you read these paths from environment variables, with a hard-coded fallback, rather than directly hard-coding them, as they are customizable.

plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
```diff
@@ -0,0 +1,161 @@
...
+// https://askubuntu.com/questions/14535/whats-the-local-folder-for-in-my-home-directory
+
+// Returns the name of the executable, without the .exe extension, or an empty
+// string on failure.
+std::string GetExecutableName() {
+  const size_t bufSize = PATH_MAX + 1;
+  char dirNameBuffer[bufSize];
+  // Read the symbolic link '/proc/self/exe'.
+  const char *linkName = "/proc/self/exe";
+  const int ret = int(readlink(linkName, dirNameBuffer, bufSize - 1));
```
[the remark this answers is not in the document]
TimWhiting: I didn't see the main.cc doing the same thing. I'll fix that.

TimWhiting:
I'm fixing all of those, and updating the pull request. 

A related question: I was reading about how web plugins are being migrated to use a federated plugin interface. [See this article](https://medium.com/flutter/how-to-write-a-flutter-web-plugin-part-2-afdddb69ece6). I see the advantages of doing it that way since if the plugin is completely in Dart, then there is no need for serialization/deserialization, and everything can be a pure function call. Do you think that approach will be adopted for Linux as well? Because unless there are platform specific APIs (such as the Window's implementation of this plugin which uses Win32 APIs) to get the document directories, I don't see a real need for going to C++ for this. We could get environment variables and read `/proc/self/exe` all in Dart and avoid serialization and C++. The other advantage would be that the user doesn't need to edit their `linux/Makefile` to compile the C++ code.

stuartmorgan:
The platform interface model is already cross-platform, and could be used for Linux implementations; it would just be a question of updating `path_provider` to use that structure. I would imagine that if you wanted to do that refactoring the flutter/plugins maintainers would be happy to accept the patch.

> The other advantage would be that the user doesn't need to edit their linux/Makefile to compile the C++ code.

Well, that's a temporary problem; we just don't have real plugin tooling written for Linux yet.

But in general, yes, there's no reason we have to use `PlatformChannel` for plugins, and when it makes sense I'm happy to have Dart implementations instead. We've actually even talked about moving toward writing plugins like the Windows `path_provider` implemention using Dart+FFI in the future.

stuartmorgan:
This is on my radar, I just haven't had a chance to get to it again. Hopefully tomorrow.

stuartmorgan:
I have a review in progress but keep not having time to finish it. Still haven't forgotten, and I promise to get to it when I can.

TimWhiting:
> /cc @robert-ancell to take a look as well.

I see that there are now conflicting files because of the new autogenerated Makefile from flutter. Would you like me to rebase before or after the review?

stuartmorgan:
You can hold off on rebasing since there will likely be another Makefile change shortly; it might be easier to update for both at once.

plugins/flutter_plugins/path_provider_fde/linux/path_provider_plugin.cc
```diff
@@ -0,0 +1,184 @@
...
+          "Unable to create a folder for the application in the ~/.local/share "
+          "directory");
+      return;
+    }
+    flutter::EncodableValue response(response_stream.str());
+    result->Success(&response);
+  } else if (method_call.method_name().compare(
+                 "getApplicationDocumentsDirectory") == 0) {
+    // Get user defined documents directory
+    const char *xdg_documents_dir = getenv("XDG_DOCUMENTS_DIR");
```
stuartmorgan: I don't think this will work. The XDG docs say that there will be in "a config file in $XDG_CONFIG_HOME/user-dirs.dirs (XDG_CONFIG_HOME defaults to ~/.config) that applications can read to find these directories". It never says that those values will be in the environment, and on my machine they aren't. Only a few higher-level things like XDG_CONFIG_HOME. You'll need to read it out of the config file.

(The fact that you need to read and parse a file may be a tipping point to where it's easier to convert path_provider to using a platform interface and then writing it in Dart.)

testbed/lib/main.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Google LLC
+// Copyright 2020 Google LLC
```
stuartmorgan: We don't update copyrights on edit.

stuartmorgan:
It turns out path_provider has already been refactored to have a platform interface, so if you want to switch this to Dart rather than proceed with C++ that should actually be fairly straightforward.

TimWhiting:
Yeah, I think I'll go for the dart implementation and close this pull request.

stuartmorgan:
@TimWhiting After some offline discussion, we've decided to add XDG directory support to flutter/packages: https://github.com/flutter/packages/pull/117. That will make implementing this in Dart trivial, and we'll be able to use it in other places as well (e.g., shared_preferences).

I really appreciate your efforts on this, and in particular pointing us to the XDG specs as the best way to handle this for Linux.

PR #684, opened 2020-02-25
Regenerate testbed solution from tool

stuartmorgan:
Regenerates the testbed solution's plugin integration using
https://github.com/flutter/flutter/pull/51246. Other than adding the
Flutter Plugins folder, it just ends up reordering some of the
entries.


stuartmorgan:
Since this is just auto-generated, I'm going to land TBR.

PR #686, opened 2020-02-27
Use generated plugin makefile on Linux

stuartmorgan:
The Flutter tool now generates a makefile suitable for inclusion in the
main makefile that adds all the necessary plugin flags and targets.
This switches to using that new system.

The header copy step is removed from the plugins, since the
generated makefile adds the necessary include paths.

Linux plugins no longer require manual build integration, so
those instructions are removed. Since there's now no need
for instructions in most cases, this removes most of the plugin
instructions. The sections about building plugins manually
are removed since there's no evidence that people are doing
manual add-to-app experimentation at this point, and it's
not something that we're supporting enough to encourage
it with partial instructions. The result is a much simpler set
of plugin instructions, which should be more approachable.


plugins/README.md
```diff
@@ -17,18 +17,15 @@ This directory contains three types of plugins:
 ## Using Plugins
 
 Since the plugins in this repository are not intended to live here long term,
-and the `flutter` tool doesn't have plugin support on all platforms yet, these
+and the `flutter` tool's plugin support isn't finalized all platforms yet, these
```
franciscojma86: isn't finalized on*

stuartmorgan: Fixed!

PR #707, opened 2020-03-31
Implement canLaunch for linux

### LINKED ISSUE #706: Implement canLaunch for url_launcher on Linux
Originally this wasn't implemented since there's no API to check if launching will succeed, but since the docs say "xdg-open supports file, ftp, http and https URLs" we could just implement it to return true for those and false for everything else. That's not perfect (a give `file:` URL may not actually be openable), but will cover the common cases, and will avoid the problem that the doc-recommended practice ("We recommend checking which URL schemes are supported using the canLaunch method prior to calling launch.") won't always fail.

kenvandine:
Returns true for URL schemes starting with ftp, http, and https

Fixes #706

### DIFF
--- a/plugins/flutter_plugins/url_launcher_fde/linux/url_launcher_plugin.cc
+++ b/plugins/flutter_plugins/url_launcher_fde/linux/url_launcher_plugin.cc
@@ -29,9 +29,9 @@ namespace {
 using flutter::EncodableMap;
 using flutter::EncodableValue;
 
-// utility function to compare url schemes
-bool StartsWith(const std::string& s, const std::string& prefix) {
-    return s.compare(0, prefix.size(), prefix) == 0;
+// Returns true if |s| starts with |prefix|.
+bool StartsWith(const std::string &s, const std::string &prefix) {
+  return s.compare(0, prefix.size(), prefix) == 0;
 }
 
 class UrlLauncherPlugin : public flutter::Plugin {
@@ -116,9 +116,9 @@ void UrlLauncherPlugin::HandleMethodCall(
       return;
     }
 
-    flutter::EncodableValue response(StartsWith(url, "https:") ||
-        StartsWith(url, "http:") || StartsWith(url, "ftp:") ||
-        StartsWith(url, "file:"));
+    flutter::EncodableValue response(
+        StartsWith(url, "https:") || StartsWith(url, "http:") ||
+        StartsWith(url, "ftp:") || StartsWith(url, "file:"));
     result->Success(&response);
     return;
   } else {
--- a/plugins/flutter_plugins/url_launcher_fde/linux/url_launcher_plugin.cc
+++ b/plugins/flutter_plugins/url_launcher_fde/linux/url_launcher_plugin.cc
@@ -44,6 +44,11 @@ class UrlLauncherPlugin : public flutter::Plugin {
       std::unique_ptr<flutter::MethodResult<EncodableValue>> result);
 };
 
+// utility function to compare url schemes
+static bool startsWith(const std::string& s, const std::string& prefix) {
+    return s.size() >= prefix.size() && s.compare(0, prefix.size(), prefix) == 0;
+}
+
 // static
 void UrlLauncherPlugin::RegisterWithRegistrar(
     flutter::PluginRegistrar *registrar) {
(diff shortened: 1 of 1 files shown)


plugins/flutter_plugins/url_launcher_fde/linux/url_launcher_plugin.cc
```diff
@@ -44,6 +44,11 @@ class UrlLauncherPlugin : public flutter::Plugin {
       std::unique_ptr<flutter::MethodResult<EncodableValue>> result);
 };
 
+// utility function to compare url schemes
```
stuartmorgan: How about "Returns true if |s| starts with |prefix|.". (The current comment says what it's used for, not what it does.)

plugins/flutter_plugins/url_launcher_fde/linux/url_launcher_plugin.cc
```diff
@@ -44,6 +44,11 @@ class UrlLauncherPlugin : public flutter::Plugin {
...
+static bool startsWith(const std::string& s, const std::string& prefix) {
```
stuartmorgan: Nits:
- No need for `static` since it's in an anonymous namespace.
- First letter should be uppercase per C++ style guide.
- Please move up to before the class declaration, so it's not sandwiched between the class declaration and definitions.

plugins/flutter_plugins/url_launcher_fde/linux/url_launcher_plugin.cc
```diff
@@ -44,6 +44,11 @@ class UrlLauncherPlugin : public flutter::Plugin {
...
+    return s.size() >= prefix.size() && s.compare(0, prefix.size(), prefix) == 0;
```
stuartmorgan: You don't need the first half of this check; compare will return non-zero if `s` is shorter than `prefix`.

plugins/flutter_plugins/url_launcher_fde/linux/url_launcher_plugin.cc
```diff
@@ -97,6 +103,30 @@ void UrlLauncherPlugin::HandleMethodCall(
       return;
     }
     result->Success();
+  // Handle canLaunch
```
stuartmorgan: You'll need to either remove these comments, or move them to the starts of the blocks they are for; clang-format will indent this as part of the block it's in, which will be confusing. (And I find the incorrect indentation form here confusing as well, although I understand the intent.)

plugins/flutter_plugins/url_launcher_fde/linux/url_launcher_plugin.cc
```diff
@@ -97,6 +103,30 @@ void UrlLauncherPlugin::HandleMethodCall(
...
+      if (url_it != arguments.end()) {
+        url = url_it->second.StringValue();
+      }
+    }
+    if (url.empty()) {
+      result->Error("argument_error", "No URL provided");
+      return;
+    }
+
+    if (startsWith(url, "https://") || (startsWith(url, "http://")) || (startsWith(url, "ftp://"))) {
```
stuartmorgan: Since the scheme ends at the `:`, you could just use `"https:"`, `"http:"`, etc.

stuartmorgan: Why not `file:` too since xdg-open supports it? I don't think it'll be a common use case, but I don't see any reason to prevent it.

plugins/flutter_plugins/url_launcher_fde/linux/url_launcher_plugin.cc
```diff
@@ -97,6 +103,30 @@ void UrlLauncherPlugin::HandleMethodCall(
...
+      flutter::EncodableValue response(true);
+      result->Success(&response);
+      return;
+    } else {
+      flutter::EncodableValue response(false);
+      result->Success(&response);
+      return;
+    }
```
stuartmorgan: You can simplify this if/else to just:
```
flutter::EncodableValue response(startsWith(url, "https://") ||
    startsWith(url, "http://") || startsWith(url, "ftp://"));
result->Success(&response);
```

stuartmorgan:
As an aside, we should definitely convert this to a pure Dart implementation at some point. Now that such a thing is possible, there's no need to use C++ here at all.

kenvandine:
I think I've addressed everything, except the CLA which I've requested to get added to our corp agreement.

kenvandine:
@googlebot I signed it!

PR #711, opened 2020-04-07
Use PluginRegistrarManager in C++ plugins

### LINKED ISSUE #704: Update C++ plugins to be multi-instantiation-friendly
The plugin template was updated to use a map of registrars so that registering with multiple engines in the same process would work, but the plugins predating that were never updated. They should all be using that construction instead.

stuartmorgan:
Updates all Windows and Linux plugins to use the new PluginRegistrarManager, as
in the updated plugin template.

Fixes #704


stuartmorgan:
Essentially mechanical change from updated template, so landing without review.

PR #712, opened 2020-04-08
Add a template for the new Linux shell

robert-ancell:
This PR is just a quick dump of a minimal application that makes use of the new GTK based Linux Flutter shell. To compile you need to include the GTK headers and link against GTK, which is normally done by using the compiler flags from `pkg-config --cflags --libs gtk+-3.0`.

No Makefile is included, but I guess we should take the existing one and modify that as necessary? Or are we still switching to CMake?

### DIFF
--- a/linux.tmpl/main.cc
+++ b/linux.tmpl/main.cc
@@ -1,0 +1,24 @@
+#include <gtk/gtk.h>
+
+#include <flutter_linux/flutter_linux.h>
+
+int
+main (int argc, char **argv)
+{
+    gtk_init (&argc, &argv);
+
+    GtkWidget *window = gtk_window_new (GTK_WINDOW_TOPLEVEL);
+    gtk_widget_show (window);
+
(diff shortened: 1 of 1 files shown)


stuartmorgan:
> No Makefile is included, but I guess we should take the existing one and modify that as necessary? Or are we still switching to CMake?

We are, but I don't want to gate having an easy shared way to build the GTK runner on that. What do you think of https://github.com/google/flutter-desktop-embedding/pull/713 as a solution for now? It's a very quick-and-dirty hack of the the existing Linux Makefile to make it simple to build so long as you build the exiting Linux runner first. That means there's not really anything to maintain in this build for now.

Once we're building the new library on CI, and flutter_tool can drop it into place, we can give it a stand-alone build instead (using Make or CMake depending on whether I've got the new CMake stuff landed by then).

stuartmorgan:
Closing in favor of #713 (which used this code).

PR #713, opened 2020-04-08
Very initial prototype of a GTK runner

stuartmorgan:
Currently requires building the normal Linux runner with a local engine
first, then piggy-backs on the resources set up by that build.


robert-ancell:
LGTM

PR #722, opened 2020-04-22
Added missing call to CoInitialize

smx-smx:
On Windows (1909 x64 at least), trying to use the "Open" or "Open Media" buttons in the testbed applications results in swallowed exceptions (nothing happening) or crashes

Running in debug mode reveals the root cause: Windows error 0x800401F0
A quick search shows that this happens when trying to use COM methods without a prior call to CoInitialize, to initialize the environment.

This solves the issue for me


smx-smx:
@googlebot I signed it!

stuartmorgan:
This was fixed last week in the runner template, I just forgot to push the change that updated `testbed` from the template. I've pushed it now; thanks for bringing it to my attention.

PR #732, opened 2020-05-17
Gtk shell plugins

robert-ancell:


plugins/color_panel/linux/include/color_panel_plugin.h
```diff
@@ -16,23 +16,19 @@
...
 #endif
 
-#if defined(__cplusplus)
-extern "C" {
-#endif
-
-FLUTTER_PLUGIN_EXPORT void ColorPanelPluginRegisterWithRegistrar(
-    FlutterDesktopPluginRegistrarRef registrar);
+FLUTTER_PLUGIN_EXPORT void color_panel_plugin_register_with_registrar(
+    FlPluginRegistrar* registrar);
```
stuartmorgan, line 31: In a GLib-API world, we can follow the normal (on other platforms) pattern of having the plugin class be public. I didn't do that here only because they were C++, so couldn't be exposed.

plugins/color_panel/linux/include/color_panel_plugin.h
```diff
@@ -16,23 +16,19 @@
```
stuartmorgan, line 16: We should probably move to `include/color_panel/color_panel_plugin.h`. I didn't do that before because it seemed pointless since it was hard to imagine there ever being more than this one function exposed in the C API, but with an object-based C API there's no reason someone couldn't have multiple public headers, so using the normal header "namespacing" rules makes sense.

plugins/color_panel/linux/include/color_panel_plugin.h
robert-ancell: Public as in call `color_panel_new()` directly? This is what made more sense to me, but I wanted to start matching the existing API as close as possible.

The registrar feels like a very complex way of doing this - it would be nicer to just have FlPlugin have methods that access everything the registrar provides.

stuartmorgan: > Public as in call `color_panel_new()` directly? This is what made more sense to me, but I wanted to start matching the existing API as close as possible.

Public meaning the plugin call would be exposed in the header, instead of completely hidden. There would still be a factory method based on the registrar. See, for instance, url_launcher ([Android](https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncherPlugin.java), [iOS](https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher/ios/Classes/FLTURLLauncherPlugin.h)); on other platforms the plugin object is public, and registration is a static method on that object.

Having the class that implements the plugin interface be public in the header would be consistent with everything but the C++ implementation (where it was impossible), so I'd prefer to do it here too.

> The registrar feels like a very complex way of doing this - it would be nicer to just have FlPlugin have methods that access everything the registrar provides.

The registrars on mobile have more capabilities than the desktop registrars currently have. I expect that desktop will expand over time to add those features.

Also, conceptually, the things the registrar provides are external details being vended to plugins, not properties of the plugins themselves (e.g., each plugin doesn't have a window, each plugin can ask for the window of the thing they are all plugged into), so composition is a cleaner fit than inheritance.

robert-ancell:
I've split out the Color Panel plugin into https://github.com/google/flutter-desktop-embedding/pull/744.

PR #733, opened 2020-05-17
Set the window size and title from configuration file

robert-ancell:


stuartmorgan:
Seems like a good way to ease the transition from the GLFW runner. We should copy the relevant files into the gtk/ directory so that it's stand-alone though, so we're building something self-contained that we can just drop in place of the existing template when it's ready to go.

PR #735, opened 2020-05-18
Add preliminary CMake build for GTK

stuartmorgan:
Adds a build for the in-progress GTK runner using the new CMake tooling,
as well as the new unpacking of GTK artifacts. Unlike the previous
Make-based implementation, this is stand-alone, rather than requiring
bootstrapping with the GLFW build first. Also doesn't require using a
local engine build.

Plugins are not yet supported, but TODOs are left for wiring them in.


stuartmorgan:
This required very few changes from the GLFW CMake build. Mostly removing the wrapper rules :)

testbed/gtk/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+endfunction()
+
+set(FLUTTER_MANAGED_DIR "${CMAKE_CURRENT_SOURCE_DIR}/flutter")
+
+# Flutter library and tool build rules.
+add_subdirectory(${FLUTTER_MANAGED_DIR})
+
+# System-level dependencies.
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(GTK REQUIRED IMPORTED_TARGET gtk+-3.0)
```
robert-ancell, line 31: We should probably add `gio-2.0` here as well as that included in a number of headers - it probably just works because GTK depends on gio.

robert-ancell, line 31: We should probably add `gio-2.0` here as well as that included in a number of headers - it probably just works because GTK depends on gio.

robert-ancell: We should also add `glib-2.0` here for completeness.

stuartmorgan: I added them to the Flutter interface library as public requirement, rather than here, since that's where the dependencies are coming from (only GTK is used directly in main.cc currently). I also added GTK there, which I was missing and should be there for completeness.

PR #739, opened 2020-05-27
Update Linux plugin build for template changes

stuartmorgan:
The linux template build files were updated to use hidden as the default
visibity. This updates all plugins, and testbed, to match.

Fixes undefined behavior in testbed (commonly, crash on launch) on Linux
due to unintended sharing of the plugin manager across plugins.


stuartmorgan:
Landing unreviewed since it's just applying a template change.

PR #740, opened 2020-05-28
Add GTK plugin loading functions

robert-ancell:
This is now supported in the Flutter engine.


testbed/gtk/flutter/generated_plugin_registrant.cc
```diff
@@ -0,0 +1,8 @@
+//
+//  Generated file. Do not edit.
```
stuartmorgan, line 2: For now let's put accurate info here, and we can restore this version later. How about:
```
// Once the Flutter tooling switches to use GTK, this will be generated automatically.
// For now, plugins must be added here manually.
```

testbed/gtk/flutter/generated_plugins.cmake
```diff
@@ -0,0 +1,7 @@
+list(APPEND FLUTTER_PLUGIN_LIST
```
stuartmorgan, line 1: Same note would apply here.

testbed/gtk/flutter/generated_plugins.cmake
```diff
@@ -0,0 +1,7 @@
...
+)
+
+foreach(plugin ${FLUTTER_PLUGIN_LIST})
+  add_subdirectory(flutter/ephemeral/.plugin_symlinks/${plugin}/linux plugins/${plugin})
```
stuartmorgan, line 5: For now this will need to be `gtk`, not `linux`, since the plan is to land the parallel implementations in `gtk` and then move them.

PR #742, opened 2020-06-01
Update Linux testbed from template

stuartmorgan:
Picks up release mode support, and some other minor template changes.


stuartmorgan:
Landing without review as it's a mechanical update from the template.

PR #743, opened 2020-06-01
Update Windows testbed from template

stuartmorgan:
Picks up release mode support, and other minor template changes.


stuartmorgan:
Landing without review since it's a mechanical update from the template.

PR #744, opened 2020-06-02
Add GTK color_panel plugin

robert-ancell:


plugins/color_panel/gtk/color_panel_plugin.cc
```diff
@@ -0,0 +1,160 @@
...
+const char kColorSelectedCallbackMethod[] = "ColorPanel.ColorSelectedCallback";
+const char kClosedCallbackMethod[] = "ColorPanel.ClosedCallback";
+const char kColorComponentAlphaKey[] = "alpha";
+const char kColorComponentRedKey[] = "red";
+const char kColorComponentGreenKey[] = "green";
+const char kColorComponentBlueKey[] = "blue";
+
+static constexpr char kWindowTitle[] = "Flutter Color Picker";
+
+G_DECLARE_FINAL_TYPE(FlColorPanelPlugin, fl_color_panel_plugin, FL,
```
stuartmorgan, line 33: I think we should put this in the header, per comments on the initial PR, even if there's nothing public on it at this point.

testbed/gtk/flutter/gtk_plugin_registrant.cc
```diff
@@ -0,0 +1,12 @@
...
+// automatically. For now, plugins must be added here manually.
+
+#include "gtk_plugin_registrant.h"
+
+#include <color_panel_plugin.h>
+
+void fl_register_plugins(FlPluginRegistry* registry) {
+  g_autoptr(FlPluginRegistrar) color_panel_registrar =
+      fl_plugin_registry_get_registrar_for_plugin(registry, "ColorPanelPlugin");
+  color_panel_plugin_register_with_registrar(color_panel_registrar);
```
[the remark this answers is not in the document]
robert-ancell: Yeah, it has to be two lines for g_autoptr to work.

testbed/gtk/flutter/gtk_plugin_registrant.cc
```diff
@@ -3,9 +3,12 @@
...
 void fl_register_plugins(FlPluginRegistry* registry) {
-  // Example plugin code:
-  // g_autoptr(FlPluginRegistrar) {name}_registrar =
-  // fl_plugin_registry_get_registrar_for_plugin(registry, "{class}");
-  //{name}_panel_plugin_register_with_registrar({name}_registrar);
+  g_autoptr(FlPluginRegistrar) color_panel_registrar =
+      fl_plugin_registry_get_registrar_for_plugin(registry, "ColorPanelPlugin");
+  FlColorPanelPlugin* color_panel_plugin =
+      fl_color_panel_plugin_new(color_panel_registrar);
+  g_object_unref(color_panel_plugin);
```
robert-ancell, line 13: Since the constructor is now public `color_panel_plugin_register_with_registrar()` seems redundant, so I just did the above. Does this match what you expected?

stuartmorgan: I would much rather follow the pattern used on all the other platforms, which is a factory method. I don't see a good reason to diverge from the other platforms on this.

robert-ancell:
@stuartmorgan I don't have commit access to this repo, but it should be ready to land.

plugins/color_panel/gtk/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+cmake_minimum_required(VERSION 3.10)
+set(PROJECT_NAME "color_panel")
+project(${PROJECT_NAME} LANGUAGES CXX)
+
+set(PLUGIN_NAME "${PROJECT_NAME}_plugin")
+
+
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(GTK REQUIRED IMPORTED_TARGET gtk+-3.0)
```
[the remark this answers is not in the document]
stuartmorgan: Sorry, I think this comment was unclear. We should still be linking to GTK in the plugin build, since it does require GTK, I just wouldn't expect us to need these to lines creating the PkgConfig::GTK library to link to since the one from the main CMake file should be visible here.

plugins/color_panel/gtk/color_panel_plugin.cc
```diff
@@ -0,0 +1,172 @@
...
+
+  const gchar* method = fl_method_call_get_name(method_call);
+  FlValue* args = fl_method_call_get_args(method_call);
+
+  g_autoptr(FlMethodResponse) response = nullptr;
+  if (strcmp(method, kShowColorPanelMethod) == 0) {
+    response = show_color_panel(self, args);
+  } else if (strcmp(method, kHideColorPanelMethod) == 0) {
+    response = hide_color_panel(self);
+  } else
```
stuartmorgan, line 129: Nit: braces.

PR #745, opened 2020-06-02
Add GTK file_chooser plugin 

robert-ancell:


stuartmorgan:
Did you want me to review these two draft PRs, or do they need more changes first?

robert-ancell:
> Did you want me to review these two draft PRs, or do they need more changes first?

I haven't been able to test this yet as a recent Flutter upgrade broke my test program but this PR is basically done. The window_size plugin needs a bit more work.

robert-ancell:
The existing Linux plugin has `registrar->EnableInputBlockingForChannel(kChannelName)` - what is the reason for this function (is it about making the dialog modal)? Do we need an equivalent here?

stuartmorgan:
> The existing Linux plugin has `registrar->EnableInputBlockingForChannel(kChannelName)` - what is the reason for this function (is it about making the dialog modal)?

IIRC the issue was that the main GLFW window would keep handling mouse input while the dialog was up, so you could accidentally trigger UI elements even though the GTK dialog acted modal.

> Do we need an equivalent here?

Hopefully not :)

robert-ancell:
@stuartmorgan this is ready for review. I haven't resolved my issues with the current Flutter build not working locally so I haven't yet fully tested it.

robert-ancell:
The existing code uses deprecated monitor querying APIs `gdk_screen_get_monitor_geometry()` etc. The new API requires GTK 3.22. Can we depend on that? Ubuntu uses GTK >3.22 since 18.04 LTS. Flutter treats the deprecation warning as errors, so we'll have to disable them to make it work in newer versions.

stuartmorgan:
> The existing code uses deprecated monitor querying APIs `gdk_screen_get_monitor_geometry()` etc. The new API requires GTK 3.22. Can we depend on that? Ubuntu uses GTK >3.22 since 18.04 LTS. Flutter treats the deprecation warning as errors, so we'll have to disable them to make it work in newer versions.

I think this comment is in the wrong PR? Per my comment there, I'm fine requiring 18.04 here (and probably for Flutter on Linux in general)

plugins/file_chooser/gtk/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+add_library(${PLUGIN_NAME} SHARED
+  "${PLUGIN_NAME}.cc"
+)
+apply_standard_settings(${PLUGIN_NAME})
+set_target_properties(${PLUGIN_NAME} PROPERTIES
+  CXX_VISIBILITY_PRESET hidden)
+target_compile_definitions(${PLUGIN_NAME} PRIVATE FLUTTER_PLUGIN_IMPL)
+target_include_directories(${PLUGIN_NAME} INTERFACE
+  "${CMAKE_CURRENT_SOURCE_DIR}/include")
+target_link_libraries(${PLUGIN_NAME} PRIVATE flutter)
```
stuartmorgan, line 16: Needs GTK.

plugins/file_chooser/gtk/file_chooser_plugin.cc
```diff
@@ -0,0 +1,198 @@
...
+const char kCanChooseDirectoriesKey[] = "canChooseDirectories";
+
+struct _FlFileChooserPlugin {
+  GObject parent_instance;
+
+  // Connection to Flutter engine.
+  FlMethodChannel* channel;
+
+  // Window to attach dialog to.
+  GtkWindow* window;
```
stuartmorgan, line 37: Same comment here as in the other PR about keeping the registrar rather than the window.

plugins/file_chooser/gtk/file_chooser_plugin.cc
```diff
@@ -0,0 +1,198 @@
...
+    for (size_t i = 0; i < fl_value_get_length(value); i++) {
+      FlValue* filter_info = fl_value_get_list_value(value, i);
+
+      if (fl_value_get_type(filter_info) != FL_VALUE_TYPE_LIST ||
+          fl_value_get_length(filter_info) != 2 ||
+          fl_value_get_type(fl_value_get_list_value(filter_info, 0)) !=
+              FL_VALUE_TYPE_STRING ||
+          fl_value_get_type(fl_value_get_list_value(filter_info, 1)) !=
+              FL_VALUE_TYPE_LIST)
+        continue;
```
stuartmorgan, line 94: We should make bad data an error instead of silently dropping it.

plugins/file_chooser/gtk/file_chooser_plugin.cc
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  value = fl_value_lookup_string(properties, kInitialFileNameKey);
+  if (fl_value_get_type(value) == FL_VALUE_TYPE_STRING) {
+    gtk_file_chooser_set_current_name(GTK_FILE_CHOOSER(dialog),
+                                      fl_value_get_string(value));
+  }
+
+  value = fl_value_lookup_string(properties, kAllowedFileTypesKey);
+  if (fl_value_get_type(value) == FL_VALUE_TYPE_LIST) {
```
stuartmorgan, line 84: This block is quite large; let's extract the loop body to a helper that takes `filter_info` and returns `filter`.

plugins/file_chooser/gtk/file_chooser_plugin.cc
```diff
@@ -0,0 +1,200 @@
...
+  FlValue* args = fl_method_call_get_args(method_call);
+
+  g_autoptr(FlMethodResponse) response = nullptr;
+  if (strcmp(method, kShowOpenPanelMethod) == 0) {
+    response = show_dialog(self, GTK_FILE_CHOOSER_ACTION_OPEN, "Open File",
+                           "_Open", args);
+  } else if (strcmp(method, kShowSavePanelMethod) == 0) {
+    response = show_dialog(self, GTK_FILE_CHOOSER_ACTION_SAVE, "Save File",
+                           "_Save", args);
+  } else
```
stuartmorgan, line 158: Nit: Needs braces.

plugins/file_chooser/gtk/file_chooser_plugin.cc
```diff
@@ -0,0 +1,201 @@
...
+      if (filter == nullptr) {
+        return FL_METHOD_RESPONSE(fl_method_error_response_new(
+            "Bad Arguments", "Allowed file types malformed", nullptr));
+      }
+      gtk_file_chooser_add_filter(GTK_FILE_CHOOSER(dialog), filter);
+    }
+  }
+
+  gint response = gtk_native_dialog_run(GTK_NATIVE_DIALOG(dialog));
+  g_autoptr(FlValue) result = fl_value_new_list();
```
stuartmorgan, line 130: Cancel needs to be null, not an empty list. Fixing now.

robert-ancell: Ah, I didn't notice that CreateResponseObject() in the old Linux plugin made a null value on an empty list.

PR #746, opened 2020-06-02
Add GTK window_size plugin

robert-ancell:


robert-ancell:
@stuartmorgan this is ready for review. I haven't resolved my issues with the current Flutter build not working locally so I haven't yet fully tested it.

stuartmorgan:
> I haven't resolved my issues with the current Flutter build not working locally

We should work through this in Discord today if you're still stuck.

plugins/window_size/gtk/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+add_library(${PLUGIN_NAME} SHARED
+  "${PLUGIN_NAME}.cc"
+)
+apply_standard_settings(${PLUGIN_NAME})
+set_target_properties(${PLUGIN_NAME} PROPERTIES
+  CXX_VISIBILITY_PRESET hidden)
+target_compile_definitions(${PLUGIN_NAME} PRIVATE FLUTTER_PLUGIN_IMPL)
+target_include_directories(${PLUGIN_NAME} INTERFACE
+  "${CMAKE_CURRENT_SOURCE_DIR}/include")
+target_link_libraries(${PLUGIN_NAME} PRIVATE flutter)
```
stuartmorgan, line 16: Should link to GTK also, per comments in the other PR.

plugins/window_size/gtk/window_size_plugin.cc
```diff
@@ -0,0 +1,203 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#include "include/window_size/window_size_plugin.h"
+
+#include <flutter_linux/flutter_linux.h>
+#include <gtk/gtk.h>
+
+// TODO(robert-ancell): Use GdkMonitor when we can depend on GTK 3.22
```
stuartmorgan, line 19: Should we just go ahead and do this now? I'm not sure we need to support below 18.04 in general, which looks like it has 3.22, and certainly for an experimental plugin I'm fine with 18.04+.

plugins/window_size/gtk/window_size_plugin.cc
```diff
@@ -0,0 +1,203 @@
...
+const char kScaleFactorKey[] = "scaleFactor";
+const char kScreenKey[] = "screen";
+
+struct _FlWindowSizePlugin {
+  GObject parent_instance;
+
+  // Connection to Flutter engine.
+  FlMethodChannel* channel;
+
+  // Window to attach dialog to.
```
stuartmorgan, line 38: Copypasta

plugins/window_size/gtk/window_size_plugin.cc
```diff
@@ -0,0 +1,203 @@
...
+
+  g_autoptr(FlStandardMethodCodec) codec = fl_standard_method_codec_new();
+  self->channel =
+      fl_method_channel_new(fl_plugin_registrar_get_messenger(registrar),
+                            kChannelName, FL_METHOD_CODEC(codec));
+  fl_method_channel_set_method_call_handler(self->channel, method_call_cb,
+                                            g_object_ref(self), g_object_unref);
+  FlView* view = fl_plugin_registrar_get_view(registrar);
+  self->window =
+      GTK_WINDOW(g_object_ref(gtk_widget_get_parent(GTK_WIDGET(view))));
```
stuartmorgan, line 195: This pattern bit me a couple of times on macOS; it works now, but with refactoring of the view/engine interaction it may become the case later that the view is null at registration time, and set later. I ended up having to change the plugins to keep the registrar, and ask for the view/window when handling messages. We should do that up front here to avoid the same breakage I had on macOS.

plugins/window_size/gtk/window_size_plugin.cc
robert-ancell: I think it's a reasonable minimum dependency - I was just copying the existing code.

plugins/window_size/gtk/window_size_plugin.cc
robert-ancell: I wasn't sure if the Registrar should be long lived, but it will be a lot simpler to keep it around.

plugins/window_size/gtk/window_size_plugin.cc
stuartmorgan: IIRC we were looking at 16.04 as a reasonable baseline at the time the initial Linux work happened, but time marches on :)

plugins/window_size/gtk/window_size_plugin.cc
```diff
@@ -0,0 +1,233 @@
...
+  g_autoptr(FlMethodResponse) response = nullptr;
+  if (strcmp(method, kGetScreenListMethod) == 0) {
+    response = get_screen_list(self);
+  } else if (strcmp(method, kGetWindowInfoMethod) == 0) {
+    response = get_window_info(self);
+  } else if (strcmp(method, kSetWindowFrameMethod) == 0) {
+    response = set_window_frame(self, args);
+  } else if (strcmp(method, kSetWindowTitleMethod) == 0) {
+    response = set_window_title(self, args);
+  } else
```
stuartmorgan, line 191: Braces.

stuartmorgan:
As a follow-up, it would be good to implement https://github.com/google/flutter-desktop-embedding/issues/483 for Linux in the GTK plugin. I never got to it for GLFW, but hopefully it would be straightforward with GTK.

stuartmorgan:
Per chat the centering isn't actually working for me, despite the values looking good on the Dart side, but the really bad issues are resolved so landing and we can follow up on the behavior later.

PR #748, opened 2020-06-03
Use GtkApplication

robert-ancell:
This replaces the simplistic minimal GTK app we had before. I will also allow us to easily add a menubar.


robert-ancell:
This will allow the menubar functionality to be easily added.

PR #755, opened 2020-06-10
Add GTK url_launcher plugin

robert-ancell:
Part of https://github.com/flutter/flutter/issues/41721

### DIFF
--- a/plugins/flutter_plugins/url_launcher_fde/gtk/url_launcher_plugin.cc
+++ b/plugins/flutter_plugins/url_launcher_fde/gtk/url_launcher_plugin.cc
@@ -83,9 +83,14 @@ static FlMethodResponse* launch(FlUrlLauncherPlugin* self, FlValue* args) {
   }
 
   FlView* view = fl_plugin_registrar_get_view(self->registrar);
-  GtkWindow* window = GTK_WINDOW(gtk_widget_get_toplevel(GTK_WIDGET(view)));
-
-  if (!gtk_show_uri_on_window(window, url, GDK_CURRENT_TIME, &error)) {
+  gboolean launched;
+  if (view != nullptr) {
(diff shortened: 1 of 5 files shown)


plugins/flutter_plugins/url_launcher_fde/gtk/url_launcher_plugin.cc
```diff
@@ -0,0 +1,152 @@
...
+  g_autoptr(GError) error = nullptr;
+  g_autofree gchar* url = get_url(args, &error);
+  if (url == nullptr) {
+    return FL_METHOD_RESPONSE(fl_method_error_response_new(
+        kBadArgumentsError, error->message, nullptr));
+  }
+
+  gboolean is_launchable = FALSE;
+  g_autofree gchar* scheme = g_uri_parse_scheme(url);
+  if (scheme != nullptr) {
```
stuartmorgan, line 64: If there's no scheme, it wasn't a valid URI, so failure of this condition should be respond with an error rather than `success(false)`.

plugins/flutter_plugins/url_launcher_fde/gtk/url_launcher_plugin.cc
```diff
@@ -0,0 +1,152 @@
...
+  g_autofree gchar* url = get_url(args, &error);
+  if (url == nullptr) {
+    return FL_METHOD_RESPONSE(fl_method_error_response_new(
+        kBadArgumentsError, error->message, nullptr));
+  }
+
+  FlView* view = fl_plugin_registrar_get_view(self->registrar);
+  GtkWindow* window = GTK_WINDOW(gtk_widget_get_toplevel(GTK_WIDGET(view)));
+
+  if (!gtk_show_uri_on_window(window, url, GDK_CURRENT_TIME, &error)) {
```
stuartmorgan, line 86: Is it valid to call this with a null window? If not, there should be a fallback to the other API since there's no reason launching a URL should fail in headless mode.

PR #760, opened 2020-06-14
Support headless mode

robert-ancell:
Associated engine changes in https://github.com/flutter/engine/pull/19038


testbed/gtk/main.cc
```diff
@@ -1,6 +1,13 @@
 #include "fl_application.h"
+#include "fl_headless_application.h"
 
 int main(int argc, char** argv) {
-  g_autoptr(FlApplication) app = fl_application_new();
-  return g_application_run(G_APPLICATION(app), argc, argv);
+  g_autoptr(GApplication) app = nullptr;
+
+  if (gdk_init_check(&argc, &argv))
```
stuartmorgan, line 7: I don't think we need the `testbed/` changes; the code I added to the GLFW Linux runner didn't actually work out, and the only reason I didn't remove them again is that I knew we'd be replacing it soon. Doing meaningful tests in headless mode doesn't turn out to be possible, so I don't see a need to support it in the template.

plugins/window_size/gtk/window_size_plugin.cc
```diff
@@ -19,6 +19,7 @@
 // See window_size_channel.dart for documentation.
 const char kChannelName[] = "flutter/windowsize";
 const char kBadArgumentsError[] = "Bad Arguments";
+const char kNoScrenError[] = "No Screen";
```
stuartmorgan, line 22: Typo: Screen

testbed/gtk/main.cc
robert-ancell: Not sure what you're referring to here? The line the comment on is the check that determines if we're running inside a graphical environment or not. Making a GtkApplication without this check will hit an assertion.

stuartmorgan: Right, I'm saying that I don't think the template actually needs to support trying to run a GTK application without a graphical environment. I tried to support that with the GLFW version thinking it would be useful for CI, and it wasn't for a variety of reasons.

robert-ancell: Discussed more on Discord, removed now.

PR #762, opened 2020-06-16
Update for GTK switch

stuartmorgan:
- Replaces all the linux/ folders with the gtk/ versions
- Resyncs the runner with the template to get the latest changes
- Minor plugin fixes.


stuartmorgan:
Landing without review; as it's essentially all mechanical changes, and it fixes breakage now that the switch has happened.

PR #782, opened 2020-08-17
Add canLaunch to Windows url_launcher

stuartmorgan:
Checks the registry for a url protocol entry for the given scheme when
checking to see if a URL can be launched.

This gives better parity with the other platforms, and means that the
url_launcher-documentation-recommended pattern of checking canLaunch
before calling launch won't fail on Windows.

### DIFF
--- a/plugins/flutter_plugins/url_launcher_fde/README.md
+++ b/plugins/flutter_plugins/url_launcher_fde/README.md
@@ -11,8 +11,3 @@ this plugin is and how to use it.
 - [x] Windows
 
 macOS and Linux are already supported by `url_launcher`.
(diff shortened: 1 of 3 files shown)


stuartmorgan:
@csells If you have a chance to review this instead of James, that would be great. (I can't request a review since apparently I hadn't made you a collaborator).

csells:
That is a way of determining "canLaunch" that I wouldn't have thought about, but you're right that Windows does list the supported schemes at the top level of HKCR and that those keys are what the shell uses to turn "open" an on URL into the right app opened with the right args. lgtm.

PR #784, opened 2020-08-22
[windows] Add min/max window size support

stuartmorgan:
Implements the window_size functionality to set minimum and maximum sizes for Windows.


plugins/window_size/windows/window_size_plugin.cpp
```diff
@@ -47,6 +46,14 @@ const char kScreenKey[] = "screen";
 
 const double kBaseDpi = 96.0;
 
+// Returns a POINT corresponding to channel representation of a size.
+POINT GetPointForPlatformChannelRepresentationSize(const EncodableList &size) {
+  POINT point = {};
+  point.x = static_cast<LONG>(std::get<double>(size[0]));
```
clarkezone, line 52: Any chance that this array could ever be empty?

stuartmorgan: It's checked at the call sites.

PR #786, opened 2020-08-24
Update Windows plugins to use references for MethodResult

stuartmorgan:
The pointer version is deprecated.


stuartmorgan:
Landing without review since it's largely mechanical, and avoids later breakage.

PR #802, opened 2020-10-10
Implemented GitHub Actions badge

leozz37:
### GitHub Actions

Implemented the CI workflow badge to README.md

CI


stuartmorgan:
Commits here are infrequent, and we don't leave the tree broken; I don't see any value in adding this badge.

PR #807, opened 2020-11-09
Migrate everything to NNBD

stuartmorgan:
Makes all plugins, and the testbed, runnable in non-null mode.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -2,10 +2,6 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-analyzer:
-  enable-experiment:
-  - non-nullable
-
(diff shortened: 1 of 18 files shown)


analysis_options.yaml
```diff
@@ -2,6 +2,10 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
+analyzer:
```
jonahwilliams, line 5: You don't need the analyzer experiment if your SDK constraints are >=2.12 , though I've noticed you need to restart any IDE/editors when that changes

jonahwilliams (approved):
LGTM with nit

analysis_options.yaml
stuartmorgan: Thanks, this was left over from when I did the bulk of this PR (against 2.11.something)

PR #815, opened 2020-11-25
Add desktop implementations of file_selector

stuartmorgan:
Adds desktop implementations of the file_selector plugin. These will temporarily be hosted here, rather than flutter/plugins, because we don't have a testing strategy for native UI in desktop plugins, so don't meet the testing requirements for flutter/plugins (whereas we already have file_chooser here without tests, so this leaves us no worse off).

While this looks like a lot of new code, there's not actually much that's new:
- The plugin code is almost a direct copy of file_chooser. Changes:
  - The classes are renamed
  - The details of the method channel API have changed slightly (different method and dictionary key names, some methods return a single item rather than everything returning a list, type filters have a different serialization structure)
  - macOS now supports UTIs in addition to extensions
  - Linux now supports mime types in addition to extensions
  - Fixed an overrelease issue in the Linux implementation of filter construction
- The example app is just a copy of the example that was landed in the app-facing file_selector plugin, with very minor changes:
  - Changed to use file_selector_platform_interface, rather than the app-facing package, since the interface should be more stable
  - Generated the Windows, macOS, and Linux projects with `flutter create .`

This initial implementation deliberately uses file_chooser code as much as possible to make for a simple transition to file_selector. Longer term we should consider:
- Rewriting the macOS plugin in Swift
- Rewriting the Windows plugin in Dart with the win32 package.


stuartmorgan:
Unfortunately despite the native plugin code being largely identical to file_chooser, the class renaming was apparently enough delta that git isn't picking it up as a copy+modification.

stuartmorgan:
(Once this lands I'll publish these plugins, under my account for now; once that's done and file_selector itself is published, I'll remove file_chooser.)

plugins/file_selector/file_selector_linux/linux/file_selector_plugin.cc
```diff
@@ -0,0 +1,238 @@
...
+  if (fl_value_get_type(value) != FL_VALUE_TYPE_MAP) {
+    return nullptr;
+  }
+
+  g_autoptr(GtkFileFilter) filter = gtk_file_filter_new();
+
+  FlValue* label = fl_value_lookup_string(value, kTypeGroupLabelKey);
+  if (label != nullptr && fl_value_get_type(label) == FL_VALUE_TYPE_STRING) {
+    const gchar* name = fl_value_get_string(label);
+    gtk_file_filter_set_name(filter, name);
```
robert-ancell, line 54: Nit: could simplify to:
`gtk_file_filter_set_name(filter, fl_value_get_string(label));`

plugins/file_selector/file_selector_linux/linux/file_selector_plugin.cc
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  bool has_filter = false;
+  FlValue* extensions = fl_value_lookup_string(value, kTypeGroupExtensionsKey);
+  if (extensions != nullptr &&
+      fl_value_get_type(extensions) == FL_VALUE_TYPE_LIST) {
+    for (size_t j = 0; j < fl_value_get_length(extensions); j++) {
```
robert-ancell, line 61: Nit: should probably use `i` instead of `j` here - it's implying a nested iteration.

plugins/file_selector/file_selector_linux/linux/file_selector_plugin.cc
```diff
@@ -0,0 +1,238 @@
...
+      g_autofree gchar* pattern =
+          g_strdup_printf("*.%s", fl_value_get_string(v));
+      gtk_file_filter_add_pattern(filter, pattern);
+      has_filter = true;
+    }
+  }
+  FlValue* mime_types = fl_value_lookup_string(value, kTypeGroupMimeTypesKey);
+  if (mime_types != nullptr &&
+      fl_value_get_type(mime_types) == FL_VALUE_TYPE_LIST) {
+    for (size_t j = 0; j < fl_value_get_length(mime_types); j++) {
```
robert-ancell, line 74: Nit: should probably use `i` instead of `j` here - it's implying a nested iteration.

robert-ancell (approved):
LGTM.

I've noted some pretty pedantic details, they're not must fixes but I figured I'd mention them.

plugins/file_selector/file_selector_linux/linux/file_selector_plugin.cc
stuartmorgan: Done. I noticed this while copying it for mime types and meant to come back and clean it up, but got sidetracked by the overrelease issue. I haven't gone back to the initial PR to check but I'm assuming when you originally wrote this code it was inlined, and so was nested, and then it got extracted to a helper (likely due to me asking for it 🙂 )

plugins/file_selector/file_selector_linux/linux/file_selector_plugin.cc
stuartmorgan: Done

plugins/file_selector/file_selector_linux/linux/file_selector_plugin.cc
stuartmorgan: Done.

plugins/file_selector/file_selector_linux/linux/file_selector_plugin.cc
robert-ancell: That was my assumption too.

stuartmorgan:
Since file_selector is published, and the changes here are minor, I'm going to go ahead and land this so I can publish the implementations so that people can start transitioning to file_selector. @cbracken if you want to do a post-landing review, I can address anything in a follow-up; if not, there will be another review point when it's time to move this to flutter/plugins.

PR #828, opened 2021-01-26
[window_size] Add show/hide functionality

stuartmorgan:
Adds 'setWindowVisibility' to show/hide the window.

Needs follow-up documentation; using it correctly for the resize-before-showing-window case is non-trivial on some platforms, and on macOS hiding will exit the app with the default Flutter app delegate configuration.

### DIFF
--- a/plugins/window_size/linux/window_size_plugin.cc
+++ b/plugins/window_size/linux/window_size_plugin.cc
@@ -26,6 +26,7 @@ const char kSetWindowFrameMethod[] = "setWindowFrame";
 const char kSetWindowMinimumSizeMethod[] = "setWindowMinimumSize";
 const char kSetWindowMaximumSizeMethod[] = "setWindowMaximumSize";
 const char kSetWindowTitleMethod[] = "setWindowTitle";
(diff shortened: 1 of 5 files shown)


plugins/window_size/linux/window_size_plugin.cc
```diff
@@ -259,6 +260,28 @@ static FlMethodResponse* set_window_title(FlWindowSizePlugin* self,
...
+  GtkWindow* window = get_window(self);
+  if (window == nullptr) {
+    return FL_METHOD_RESPONSE(
+        fl_method_error_response_new(kNoScreenError, nullptr, nullptr));
+  }
+  if (fl_value_get_bool(args)) {
+    gtk_widget_show(GTK_WIDGET(window));
+  } else {
+    gtk_widget_hide(GTK_WIDGET(window));
+  }
```
robert-ancell, line 280: FYI: This can be done more simply as:
```c
gtk_widget_set_visible(GTK_WIDGET(widget), fl_value_get_bool(args));
```

PR #836, opened 2021-02-23
Migrate file selector example to NNBD

stuartmorgan:
Just to make absoulety sure that there are no issues with the platform
implementations for NNBD (e.g., bad dependency versions).

Also tidies up the example for this repo's analysis_options.yaml, since
it was written against flutter/plugins originally which has different
settings, as well as replacing some deprecated material buttons.

### DIFF
--- a/plugins/file_selector/example/lib/get_directory_page.dart
+++ b/plugins/file_selector/example/lib/get_directory_page.dart
@@ -4,13 +4,13 @@ import 'package:flutter/material.dart';
 /// Screen that shows an example of getDirectoryPath
 class GetDirectoryPage extends StatelessWidget {
   void _getDirectoryPath(BuildContext context) async {
-    final String confirmButtonText = 'Choose';
-    final String directoryPath = await FileSelectorPlatform.instance.getDirectoryPath(
+    const confirmButtonText = 'Choose';
+    final directoryPath = await FileSelectorPlatform.instance.getDirectoryPath(
       confirmButtonText: confirmButtonText,
     );
     await showDialog(
       context: context,
-      builder: (context) => TextDisplay(directoryPath),
+      builder: (context) => TextDisplay(directoryPath ?? 'Unknown'),
     );
   }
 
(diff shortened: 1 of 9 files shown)


plugins/file_selector/example/lib/open_image_page.dart
```diff
@@ -6,14 +6,15 @@ import 'package:flutter/material.dart';
 /// Screen that shows an example of openFiles
 class OpenImagePage extends StatelessWidget {
   void _openImageFile(BuildContext context) async {
-    final XTypeGroup typeGroup = XTypeGroup(
+    final typeGroup = XTypeGroup(
       label: 'images',
       extensions: ['jpg', 'png'],
     );
```
cbracken, line 12: Unnecessary for this PR, but looking at XTypeGroup definition, it could be made const, which would allow this to be made const.

plugins/file_selector/example/lib/open_image_page.dart
```diff
@@ -25,15 +26,13 @@ class OpenImagePage extends StatelessWidget {
   Widget build(BuildContext context) {
     return Scaffold(
       appBar: AppBar(
-        title: Text("Open an image"),
+        title: Text('Open an image'),
```
cbracken, line 29: This could be const.

plugins/file_selector/example/lib/get_directory_page.dart
```diff
@@ -4,29 +4,27 @@ import 'package:flutter/material.dart';
...
+      builder: (context) => TextDisplay(directoryPath ?? 'Unknown'),
     );
   }
 
   @override
   Widget build(BuildContext context) {
     return Scaffold(
       appBar: AppBar(
-        title: Text("Open a text file"),
+        title: Text('Open a text file'),
```
cbracken, line 21: This could be const.

cbracken (approved):
LGTM with minor nits.

plugins/file_selector/example/lib/open_image_page.dart
stuartmorgan: Fixed all constant `Text` elements.

PR #837, opened 2021-02-25
[file_selector] Ignore empty accept group lists on macOS

stuartmorgan:
The null safety migration for file_selector defaults the accept group to
an empty set rather than null, so empty set needs to be treated as no
filter. That's already the case for Windows and Linux; this makes macOS
consistent with that.


cyanglaz (approved):
LGTM

PR #841, opened 2021-03-21
Update README.md

SushantChandla:
Updated version in README.md from 0.0.3 to 0.0.4


SushantChandla:
@googlebot I signed it!

stuartmorgan:
Thanks, but that section of the README is simply intended to be representative of what using the unendorsed plugin will look like. The exact version someone wants to require for their app is, as with any package, up to them.

PR #846, opened 2021-03-29
[window_size] Fix unconstrained minimum window size in Linux

robert-ancell:
Flutter uses (0, 0) as unconstrained, GTK uses (-1, -1).

This would mean unsetting the minimum window size from Flutter would have no effect in GTK.


robert-ancell:
Not sure if we need this change. Without it you can't get back to the default minimum size of `(-1, -1)` except by passing a negative value for the min size from Flutter (which is not documented as being valid). However, a minimum size of `(0, 0)` is really the equivalent of `(-1, -1)` as GTK won't let you resize smaller than the *requisition size* (the size GTK determines the window should be).

PR #847, opened 2021-03-29
[window_size] Fix incorrect value for unconstrained maximum size.

robert-ancell:
The previous code used (-1, -1), which means use the requistion size (the size that GTK has determined for the window) and resulted in windows that were unresizable.

Fixes https://github.com/google/flutter-desktop-embedding/issues/844


stuartmorgan (approved):
LGTM

robert-ancell:
Just updated it to do the same transform when requesting the size. We do have an interesting case if the size is `(-1, -1)` (i.e. requisition size) we can't expose that to the Dart side as Dart doesn't have that concept. Instead that will look like its unconstrained. This is probably only an issue if you have non-Dart code messing with the window size.

stuartmorgan (approved):
Still LGTM

PR #848, opened 2021-03-30
[window_size] Fix getWindowMinSize in Linux

robert-ancell:
The GTK default is (-1, -1), which looks like Size(-1, -1) to Flutter.
GTK uses -1 for requisition size (the size GTK has calculated), so we map that to (0, 0).


robert-ancell:
This fixes the issue if you make a trivial Flutter app that does:
```dart
print(await window_size.getWindowMinSize());
```
You get:
```
Size(-1.0, -1.0)
```

stuartmorgan (approved):
LGTM

PR #866, opened 2021-07-22
Fix alphabetical key shortcuts.

guibarrette:
Fix alphabetical key shortcuts. A SHIFT modifier key was registered by accident since the keyLabel from a LogicalKeyboardKey in Flutter changed to an upper case value

Fixes https://github.com/google/flutter-desktop-embedding/issues/865


guibarrette:
@googlebot I signed it!

stuartmorgan (approved):
LGTM, thanks!

PR #894, opened 2022-01-26
Remove file_selector_windows

stuartmorgan:
Removes the Windows implementation, as it has now moved to
flutter/plugins. Since the implementation is no longer part of this
repository, removes it from the shared example app as well.


cbracken (approved):
LGTM stamp from a Japanese personal seal

PR #904, opened 2022-07-11
[file_selector] Add Linux unit tests

stuartmorgan-g:
Part one of preparation to move this implementation to flutter/plugins. Unlike the previous moves, I'm adding tests before moving, as it makes reviewing the deltas easier; the final move can make only the minor changes needed for the move itself instead of having functional diffs as well.

Since the tests run in the context of the example app, and since there are no longer multiple implementations in this repo to share an example between, this moves the example app into the package.

The testing here isn't as complete as I would like; it's testing internals that cover a significant portion but definitely not all of the native logic. This is a balance between getting some meaningful testing in place (and thus allowing moving and endorsing it) against the amount of time it would take me to fully test (ramping up on doing mocking, callback wrapper patterns, etc. in GObject).

(Part two will add a Dart implementation to move the method channel in-package.)


plugins/file_selector/file_selector_linux/linux/test/file_selector_plugin_test.cc
```diff
@@ -0,0 +1,166 @@
...
+  EXPECT_EQ(gtk_file_chooser_get_action(GTK_FILE_CHOOSER(dialog)),
+            GTK_FILE_CHOOSER_ACTION_SAVE);
+  EXPECT_EQ(gtk_file_chooser_get_select_multiple(GTK_FILE_CHOOSER(dialog)),
+            false);
+  g_autofree gchar* current_name =
+      gtk_file_chooser_get_current_name(GTK_FILE_CHOOSER(dialog));
+  EXPECT_STREQ(current_name, "foo.txt");
+  // TODO(stuartmorgan): gtk_file_chooser_get_current_folder doesn't seem to
+  // return a value set by gtk_file_chooser_set_current_folder, or at least
+  // doesn't in a test context, so that's not currently validated.
```
cbracken, line 152: What does it return? Looks like it's meant to return null if it was unable to load the directory passed to `gtk_file_chooser_set_current_folder`.

cbracken (approved):

plugins/file_selector/file_selector_linux/linux/test/file_selector_plugin_test.cc
stuartmorgan: I believe it was null. My guess is that it's because the dialog hasn't actually been shown yet, so

> Note that this is the folder that the file chooser is currently displaying

is arguably saying that null is correct behavior before showing it.

PR #907, opened 2022-08-18
[file_selector] Remove Linux implementation

stuartmorgan-g:
This plugin has been moved to flutter/plugins.

Also removes the old `file_chooser` README redirect, since
`file_selector` has now existed for quite a while, so `file_chooser`
references would be old enough that they aren't worth continuing to
handle.


cbracken (approved):
LGTM stamp from a Japanese personal seal

PR #912, opened 2023-01-01
how to control window size

Mahmoud7Osman:
Because i went through a tough time searching how to use this plugin, even chatGPT doesn't know how to use it right.


stuartmorgan:
Closing, as we cannot accept PRs without a signed CLA. Please feel free to file a new PR if you sign it.

Also, information about using `window_size` should be in the `window_size` package, not the repo-level README.

PR #913, opened 2023-01-09
Replace testbed/ with plugin examples

stuartmorgan-g:
The testbed/ directory is a relic of a much earlier stage of desktop development, when maintaining apps was cumbersome (because they had to be recreated from scratch on a regular basis, before the template stabilized) and it was useful to have a desktop-enabled example handy for basic functionality like scrolling and typing that were still being implemented.

This removes testbed/, and replaces it with standard in-package examples for the two remaining plugins (the other purpose of testbed). Each is created with `flutter create -t plugin .` with the following changes:
- Removed all the template comments.
- Removed the template-standard used of `flutter_lints` (as the repo uses flutter/plugins-based lints).
- Added the copyright notice.
- Changed the example Dart code:
  - `menubar`'s example is the counter app, but with the resize-on-launch logic taken from `testbed`.
  - `testbed`'s example is testbed's code, but with the keyboard and scrolling parts removed.
- Changed the template-standard integration test to be a simple sanity check that calling the plugin works.
- Removed the (new!) template-standard macOS native unit tests, since backfilling unit tests is out of scope here. The scaffolding is still in place in case we want to add some later.

Having the standard example structure will make the plugins easier to maintain when we do need to touch them, and easier for people to find out how to use.

Tangential change:
- Updates the CI hosts since the macOS and Linux hosts that were there are deprecated.


stuartmorgan:
This looks like a huge review, but other than the Dart files everything in the new example/ directories (which is almost everything) is just boilerplate created by `flutter create`.

gspencergoog (approved):
32384589-a60f0e74-c078-11e7-9bc1-e5b5287aea9d

This looks good, thanks for cleaning it up.

PR #919, opened 2023-05-03
Default window size feature

DmitriiProshutinskii:
This feature implements on macOS, Windows and Linux method `setDefaultSize`
It helps us to make the window with size which can show more content and add flexibility to user to use this package


stuartmorgan:
Thanks for the contribution, but this plugin is not accepting new features as [documented in the README](https://github.com/google/flutter-desktop-embedding/tree/main/plugins/window_size#scope).

